PR #16, opened 2018-09-18
Update Operator branches

eric-schneider:
Updated the production and staging playbooks to reflect the latest Operator branches.


eric-schneider:
@amarantha-k Can you merge this and republish the site? There are Operator 1.1 changes that have made it into the site that shouldn't be there.

PR #18, opened 2018-10-04
add license and usage to the tabs block extension

mojavelinux:


mojavelinux:
@amarantha-k I added a license to the tabs block extension (MPL 2.0). This allows the extension to be maintained as part of the Antora project. The upstream version is located here: https://gitlab.com/antora/antora-asciidoctor-extensions/tree/master/tabs-block

Once we figure out how to distribute the extension as a package, we can reconfigure the playbook repository to consume this extension as a dependency. This is just the first step in that process.

PR #21, opened 2018-10-09
DOC-4229 : Remove pre-release label

amarantha-k:
Note: Merge for 6.0 GA


mojavelinux:
@amarantha-k to remove the "Developer Preview" label from the component menu in the navbar, you'll need to modify the UI template (https://github.com/couchbase/docs-ui/blob/master/src/partials/header-content.hbs#L23), commit it to master, then update the UI bundle URL in the playbook to use the new version.

The component menu is one area of the content that is managed in part by the UI (though we can continue to work to remove that coupling).

PR #22, opened 2018-10-09
add tutorials to playbooks

jamiltz:


jamiltz:
@amarantha-k this PR is ready to be tested/merged.

By use `tutorials/` in the component name prefix (https://github.com/couchbaselabs/hotel-finder-react-native/commit/2d4c4e48313edfa52948568a9962d6d1a758240c), it will overwrite the URL of tutorials that are already deployed (for example https://docs.couchbase.com/tutorials/hotel-finder/ios.html) which means we don't need to put redirects in place.

mojavelinux:
Hmm, I strongly recommend using a slash in the component name. Antora was not designed with that in mind, so it's not good to rely on that behavior (which is essentially a side effect). It would be much better to use "tutorials" as the component name and a module for each tutorial.

mojavelinux:
At the same time, we could study this use case to try to understand if more flexibility is needed in Antora. But I strongly recommend staying without the boundaries of the specified Antora behavior so that you don't get surprised by changes.

jamiltz:
@mojavelinux Thanks for the review!

> It would be much better to use "tutorials" as the component name and put each tutorial in a module.

Had considered that but we'd like to keep docs with code, and for code to be in different repos. And using a single component with multiple repos doesn't seem to be recommended?

mojavelinux:
On the contrary, that's exactly how Antora was designed. You can absolutely have modules for the same component in different repositories. In fact, that feature is already used in several places in the Couchbase docs. It's really the whole reason modules exist.

jamiltz:
Do we have an example of that in server docs?

mojavelinux:
First, see https://docs.couchbase.com/home/contribute/component-configuration.html

The analytics is set up this way. See https://github.com/couchbase/docs-analytics (it becomes server/analytics in the URL)

jamiltz:
Will do. Thanks @mojavelinux !!

jamiltz:
Updated the change set per Dan's feedback.

mojavelinux:
Almost. Now I see that the tutorials are in a topic folder inside the ROOT module. What I meant to suggest is to put the tutorials in a module with the name of the tutorial. So, for example, hotel finder would look like this:

```
antora.yml
modules/
  hotel-finder/
    pages/
      android.adoc
      ios.adoc
```

The resulting URL will then be: tutorials/hotel-finder/android.html

jamiltz:
Got it. Will update.

mojavelinux:
The hotel finder now looks right. I noticed in the todo tutorial that there are a lot of pages directly under modules. Are those just sitting there temporarily until  you have a chance to move them into place?

jamiltz:
Yes exactly.

jamiltz:
@amarantha-k this PR is ready for merging. I have intentionally only updated the staging playbook to check the result on staging only for now.

mojavelinux:
Congratulations!

PR #24, opened 2018-10-11
DOC-4235

RichardSmedley:


amarantha-k:
While the change is not in this PR, I see that the release notes include install instructions. Shouldn't the install instructions be under "Start using the SDK"?

RichardSmedley:
Looks like a more recent PR has pulled in v185 of the ui, which supercedes this - no need to merge now.

PR #25, opened 2018-10-17
Gulpfile with livereload

jamiltz:


jamiltz:
Keeping the PR open for the doc team to try it out.
@mojavelinux a review would be much appreciated on this one.

mojavelinux:
I like where you're going with this. At the same time, I wish I could say Antora provided this so you didn't have to develop this script. In that light, I've opened an issue in Antora to get the ball rolling on that. See https://gitlab.com/antora/antora/issues/329.

I'm a bit on the fence whether it makes sense to invoke the generator directly or invoke it via the CLI. The benefit for invoking it via the CLI is that it's a separate process and any residual memory will get cleared after each run. On the other hand, if you use the API, it's possible to implement a custom pipeline that might do an incremental build.

Suffice to say, Antora is an open platform for a reason. We encourage you to pull it apart, play around with it, and find interesting things to do with it like this.

jamiltz:
Ok thanks @mojavelinux. Added comments to keep that in mind.

Will keep the PR open until there's confirmation it works for everyone on the team.

PR #52, opened 2019-01-23
Dockerize the docsearch-scraper CI job

mojavelinux:
Leverage the Docker support in Jenkins to drastically simplify the build script that runs the docsearch scraper.


mojavelinux:
@jamiltz Thanks again for enabling Docker support on Jenkins. That change has had a huge impact on the simplicity of the jobs, as this PR shows.

PR #56, opened 2019-02-06
Update backup repo branch to use alice

amarantha-k:


amarantha-k:
Dan, the build failed when I updated the branch to use "alice" instead of "master". Just wanted to ask if you see anything amiss. I'll investigate in the am.

mojavelinux:
The reason the build failed is because the version number in antora.yml in the vulcan and alice branch are both 5.5. The version number in the alice branch should be 6.0 (as far as I understand it).

PR #64, opened 2019-03-15
Build for new version

amarantha-k:


mojavelinux:
One thing you need to be mindful of is that when you introduce a new version of server, you need to add a branch for every repository that contributes to the server component. What might be helpful is to use a line comment to mark these entries in the playbook. That way, when upgrading, you don't forget one of them.

I believe you've caught all of them in this change. However, I would itemize the branches for analytics to make it clearer which branches are being used, since it's one of the repositories that contributes to server.

staging-antora-playbook.yml
```diff
@@ -81,7 +84,7 @@ asciidoc:
   - ./lib/tabs-block.js
 ui:
   bundle:
-    url: https://github.com/couchbase/docs-ui/releases/download/v201/ui-bundle.zip
+    url: https://github.com/couchbase/docs-ui/releases/download/v202/ui-bundle.zip
```
jamiltz:
```suggestion
    url: https://github.com/couchbase/docs-ui/releases/download/v201/ui-bundle.zip
```

I suggest bumping the version as part of https://github.com/couchbase/docs-site/pull/61. Otherwise the Scala item on the top nav will 404.

PR #68, opened 2019-04-09
Update CI pipeline to use managed infrastructure

mojavelinux:
This PR should only be merged **after** docs.couchbase.com has been switched over the to new production infrastructure (which incorporates nginx). The pipeline retrieves information about the infrastructure (such as the CloudFront distribution ID) from an S3 bucket (placed there by Terraform), then uses that information to issue the AWS commands. It also pushes out nginx rewrite rules (a combination of manual rules from this repository and rules generated by Antora) to the nginx server and reloads it.


mojavelinux:
This was merged into master manually.

PR #69, opened 2019-04-09
Added more examples of linking within a module

eric-schneider:
It's important to have examples that show how to link within a module that has a nested directory structure, especially since several of our modules are organized this way.


jamiltz:
Great addition!

PR #84, opened 2019-05-10
Opening a nav link in a new tab

jamiltz:


jamiltz:
Works w/o quotes. I removed them 👍

PR #98, opened 2019-07-09
DOC-5183 Add test redirect for .NET SDK API

amarantha-k:


mojavelinux:
Seems reasonable to me. I might suggest dropping the `-current` suffix so it goes:

/sdk-api/couchbase-net-client/ -> /sdk-api/couchbase-net-client-2.7.8/

Seems more intuitive for a user to type. And since the API docs don't really care about the URL path, it's safe to do it this way.

amarantha-k:
@RichardSmedley Please take a look at the updates and let me know if you have any reservations dropping the "-current" from the stable URL for SDK APIs.

RichardSmedley:
I have no reservations myself (I think it's a good idea - given the differences).
Waiting for comments from the SDK team, but we might as well go ahead with this one as a test, while we wait. :+1:

PR #99, opened 2019-07-10
enable xref validator in staging and production site

mojavelinux:
- add @antora/xref-validator to Docker image for production build
- add @antora/xref-validator to Netlify build
- configure staging build to use xref validator
- configure production build to use xref validator
- add @antora/xref-validator to authoring tools


mojavelinux:
Netlify seems to be experiencing issues at the moment, so I'm going to have to rely on the production build to verify the settings. But it will only publish if the validation succeeds, so that approach imposes no risk.

mojavelinux:
Okay, Netlify is back.

PR #101, opened 2019-07-19
Test fix for correct link -- DOC-5503

RichardSmedley:


mojavelinux:
The nginx rewrite rules don't get used on the staging site (that's a different rewrite engine). Currently, there's no way to test the nginx rules for production (though publishing will be halted if there's a syntax error).

This change looks right to me, so I recommend proceeding with the merge.

PR #106, opened 2019-08-07
Update to disable prod builds temporarily

amarantha-k:


mojavelinux:
Yep, I believe this will work. Remember that after a change to the Jenkinsfile, the build has to run once for the changes to take affect.

Another possible strategy is to remove the `--fetch` flag from the Antora command. That will rebuild the site from the latest cache. Though given we don't know the exact cache policy, that approach might be less reliable.

I'm quite surprised it's not possible to simply disable the build temporarily from the Jenkins UI. I checked and there's no such option (unlike a freestyle build).

amarantha-k:
Got it. Actually, there is a checkbox to Disable Projects under Build Triggers, but the changes are not saved. In fact, I don’t see a Save button either. This is very different from earlier versions where we could just edit the job configuration and save it. 
I’ll merge the PR and run the build. 🤞

mojavelinux:
> there is a checkbox to Disable Projects under Build Triggers, but the changes are not saved. In fact, I don’t see a Save button either. 

That's correct. The whole UI is read only when using a pipeline project (instead of a freestyle project). Jenkins does a very poor job of relying this information because it reuses the form for the view mode. It should not be a form.

amarantha-k:
The build completed and Build Periodically and Poll SCM fields are unchecked now. Thanks for reviewing the PR, Dan!

PR #109, opened 2019-08-12
Dev preview partial

jamiltz:
Follow up from @simon-dew's commit [27cad2b](https://github.com/couchbase/docs-server/commit/ab6d3eafa331bb66a04f87b28281a0837517e4a4#diff-27cad2b57b742b638d75023b6acfe0ff) which adds a partial in the `server` component.

This PR adds a slightly modified version of that partial to the `ROOT` module in the `home` component and documents a usage example.

I've removed the h2 and xref to the developer preview mode page (specific to CB Server) although we could certainly add those back in in a sub-sequent PR with their respective usage examples.

Preview: https://deploy-preview-109--cb-docs-staging.netlify.com/home/contribute/includes.html#couchbase-documentation-partials

### DIFF
--- a/home/modules/contribute/pages/includes.adoc
+++ b/home/modules/contribute/pages/includes.adoc
@@ -105,6 +105,21 @@ For example, if you tag code in a YAML file, place a hash (`+#+`) before the tag
     - public/
 ----
 
+== Couchbase Documentation Partials
+
+=== Developer Preview Partial
+
+To call out a section of a page as Developer Preview, use the `developer-preview` partial.
+Otherwise if the whole page is Developer Preview content, you can use the xref:attributes-and-roles.adoc#custom-page-attributes[page-status] attribute.
+
+----
+\include::home:ROOT:partial$developer-preview.adoc[]
+----
+
+Renders as:
+
+include::home:ROOT:partial$developer-preview.adoc[]
+
 == Learn More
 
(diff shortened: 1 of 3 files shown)


home/modules/contribute/pages/includes.adoc
```diff
@@ -105,6 +105,21 @@ For example, if you tag code in a YAML file, place a hash (`+#+`) before the tag
...
 
+== Couchbase Documentation Partials
+
+=== Developer Preview Partial
+
+To call out a section of a page as Developer Preview, use the `developer-preview` partial.
+Otherwise if the whole page is Developer Preview content, you can use the xref:attributes-and-roles.adoc#custom-page-attributes[page-status] attribute.
+
+----
+\include::home:ROOT:partial$developer-preview.adoc[]
```
RichardSmedley: Should there be a note here about the more typical PATH found in most CB components (ie, not /home/)?

jamiltz: In theory that path will be the same regardless of where the partial is included (but haven't tried it yet). The idea is to have a single partial which can be included throughout the site's components. I've just added the staging link from the PR checks on the PR's 1st comment so you can now see the output.

RichardSmedley: Okay, we'll see how it goes :+1:

jamiltz: Yes 🙂

PR #142, opened 2020-01-16
Update homepage xrefs to point to new Java pages

eric-schneider:
This change is a stopgap to paint over the the major page alias issues that are occurring in the Java SDK 3.0 docs.


eric-schneider:
@amarantha-k @RichardSmedley There are major issues in the Java SDK 3.0 docs right now with include directives and page aliases not resolving. Only merge this change once you've decided not pull the 3.0 docs. If you rebuild the site without the 3.0 docs, the xrefs in this change will break again.

PR #157, opened 2020-04-03
Explicitly mention xcode install

acoliver:
xcode isn't required to install the JDK. I at least went years without having it installed as a Java developer. It failed for me without xcode.

### DIFF
--- a/home/modules/contribute/pages/install-antora.adoc
+++ b/home/modules/contribute/pages/install-antora.adoc
@@ -12,7 +12,7 @@ If you've never used Antora, you may first need to instal Node using nvm.
 Follow the instructions below that correspond to your operating system:
 
 * {url-linux}[Linux: Install Node^]
-* {url-macos}[macOS: Install Node^]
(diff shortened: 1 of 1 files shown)


mojavelinux:
As far as I'm aware, xcode is not required to install Node.js on macOS. What method did you use to install Node.js?

I know there are a number of writers who use macOS. Can anyone confirm whether or not xcode is required? @eric-schneider @simon-dew @amarantha-k 

If it turns out xcode is required for some particular installation method, it should be documented in the Antora docs, not here.

simon-dew:
I can confirm that I installed nvm and node.js on macOS Mojave following the instructions [here](https://docs.antora.org/antora/2.3/install/macos-requirements/#install-nvm), and I did not need to install xcode.

mojavelinux:
@acoliver As far as I know, xcode is not required. If you can demonstrate that it is required, please open an issue in https://gitlab.com/antora/antora and we'll address it there.

PR #177, opened 2020-04-24
move the blue bar/logo section

robashcom:
move the blue bar down one section in page position


amarantha-k:
Closing without merge as there's an updated PR.

PR #180, opened 2020-04-27
Developer page for prod

amarantha-k:


mojavelinux:
@amarantha-k I'd like to see the styles for the developer page get moved into the UI. We really shouldn't be putting styles directly into an AsciiDoc document as that violates the separation of content and presentation.

amarantha-k:
@mojavelinux I agree and it's on my list of things to work on. Design review for this page is in progress and we'll plan to move the CSS into the UI after that.

mojavelinux:
Super!

PR #189, opened 2020-05-07
add custom site generator component to generate site manifest

mojavelinux:


mojavelinux:
@amarantha-k This PR switches the build to a custom site generator. This custom software component adds the extra step of generating the site manifest, which the archive site uses to resolve xrefs to the production site. It works with whatever version of Antora is installed.

Once this functionality is incorporated into Antora, we'll be able to drop the custom site generator. The site will still build if the generator is not specified, so the writers don't need to care about it.

PR #191, opened 2020-05-11
Update to latest UI bundle

amarantha-k:


amarantha-k:
@mojavelinux Surprised that the preview build failed when stage went through fine. Any pointers?

mojavelinux:
It was a communication error with GitLab:

```
ERR! /usr/bin/git ls-remote -h -t https://gitlab.com/antora/xref-validator.git
```

Also note that Netlify is limiting the build to 15 minutes and we're right up against that limit. So sometimes it fails because it runs out of time (not this time though).

amarantha-k:
Thanks!

amarantha-k:
Merging as build was successful but timed out during post processing phase.

PR #192, opened 2020-05-14
merge prod and archive UI

mojavelinux:


mojavelinux:
@amarantha-k All we want to do here is verify that this UI update does not affect production, specifically the view latest button.

mojavelinux:
Thanks for verifying!

PR #200, opened 2020-05-29
Update home page for archive site

amarantha-k:


amarantha-k:
@mojavelinux The CSS eventually needs to be moved to docs-ui repository.

mojavelinux:
@amarantha-k Can you use xrefs instead of absolute URLs for the links?

amarantha-k:
If that's not a blocker, I'd prefer to do that by Monday. There are some more older SDK links that need to be included on this page.

mojavelinux:
I think it's pretty important to stick with xrefs where possible. I can make that update for you, if you'd like. It probably wouldn't take me long.

I noticed "Node.js" is written as "Nodejs". To be consistent with other uses on the site, I think it should be the former.

amarantha-k:
That's great if you can update the xrefs. Thanks!

amarantha-k:
Thanks for the updates, @mojavelinux. Are we good to merge the PR?

mojavelinux:
@amarantha-k If you're happy with the result, then yes, it's ready to merge.

amarantha-k:
Looks good; +1 to merge.

mojavelinux:
Okay, merging!

PR #193, opened 2020-05-14
remove archived branches from production site

mojavelinux:

PR #201, opened 2020-06-01
Remove archived versions from master site search configuration

amarantha-k:


mojavelinux:
This should be part of #193. I've reapplied the change to that PR branch.

PR #228, opened 2020-06-30
Update landing pages and playbook file

amarantha-k:


amarantha-k:
@mojavelinux I'm waiting to confirm that the Cloud landing page builds properly. We should be good to merge this PR soon after.

mojavelinux:
@amarantha-k The only thing that seems off is that the SDKs are not actually laying out as 2 columns.

amarantha-k:
Yes, I see that. Why don't we merge this PR so you can proceed with the next steps, and I'll continue to look into fixing the SDK card layout.

mojavelinux:
Perfect. I'll click merge.

PR #230, opened 2020-06-30
fix links on cards

amarantha-k:


amarantha-k:
Does not work as expected. Closing without merge.

PR #271, opened 2020-09-15
Port updates to swagger ui macro to staging env

amarantha-k:


amarantha-k:
@mojavelinux We're unable to get any of the yaml config or rest-api files to render on the docs-staging site and suspect this is because the staging env did not have the recent updates made in the production branch.

mojavelinux:
You're correct that updating Antora required a corresponding update to the extension. In Antora 2.3, it's no longer possible to modifying header_attributes directly (it just gets overwritten). Instead, the attribute must be set on the property on the Antora virtual file.

ibsoln:
@amarantha-k great stuff that macro seems to render rest api yaml now.
Was there a similar change to the json-config-ui-block-macro.js macro? As that still does not render.

mojavelinux:
Yes, that would require the same change. You can see when this change was made to the production site here: https://github.com/couchbase/docs-site/commit/150de648607f4ceec24119c9a647679d4e1fc4f1

PR #287, opened 2020-10-01
upgrade site to Antora 3 and leverage latest version segment feature to manage current URL

mojavelinux:


mojavelinux:
@amarantha-k This is now ready for review.

For example:

https://deploy-preview-287--cb-docs-staging.netlify.app/server/6.5/introduction/why-couchbase.html

Notice that both the link and the canonical URL point to the current URL.

A request for version 6.6 will be routed to the current URL:

https://deploy-preview-287--cb-docs-staging.netlify.app/server/6.6/introduction/why-couchbase.html

And the View Latest button follows aliases:

https://deploy-preview-287--cb-docs-staging.netlify.app/c-sdk/2.10/sdk-authentication-overview.html

All of this is now handled by Antora without any use of helper functions in the UI. Here are the magic playbook settings that canonicalize the current URLs:

```
urls:
  latest_version_segment: current
  latest_version_segment_strategy: redirect:to
```

That says redirect the latest version to the latest version segment, which is current.

amarantha-k:
This is a much awaited update! 
- When requesting for a version 3.0 of the SDKs, they are not routed to the current URL.The View Latest does resolve to the current URL.

mojavelinux:
@amarantha-k Can you clarify what you mean? For me, it is requesting the current URL in that case. Maybe walk be through where you are clicking?

amarantha-k:
Sorry - this must be a caching issue at my end. Now I see that it is requesting the current URL as well.

amarantha-k (approved):
lgtm

mojavelinux:
Excellent!

Just to clarify, if a page is not present in the latest version, and there is no way to get there by following aliases, then the View Latest button will be hidden. As an example, see https://deploy-preview-287--cb-docs-staging.netlify.app/server/5.0/connectors/kafka-3.1/kafka-intro.html If and when an alias is added, it will show up again.

PR #311, opened 2020-12-15
enable 404 page with router

mojavelinux:


mojavelinux:
@amarantha-k Here's an example of this feature in action: https://deploy-preview-311--cb-docs-staging.netlify.app/operator/current/does-not-exist.html

PR #346, opened 2021-03-25
demonstrate how to make a lean site build

mojavelinux:
- define primary-site-url to point to https://docs.couchbase.com
- remove (or comment out) any sources you don't want to include in this site build


mojavelinux:
The key here is the addition of the `primary-site-url` attribute. That tells the custom generator to import the files from the manifest from that site into the content catalog to resolve xrefs. That way, you only have to build the content you want to change.

PR #348, opened 2021-04-02
roll out search enhancements, including search filter, to production

mojavelinux:
* upgrade UI that includes search filter and other search enhancements

### DIFF
--- a/antora-playbook.yml
+++ b/antora-playbook.yml
@@ -129,6 +129,6 @@ asciidoc:
   - ./lib/tabs-block.js
 ui:
   bundle:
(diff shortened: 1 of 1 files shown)


mojavelinux:
@amarantha-k You may notice that the search filter in the deploy preview works even without the updates to the search index. I designed the search filter to work with either schema. In doing so, it allows us to roll out the feature without disruption to the end user.

mojavelinux:
@amarantha-k This is ready for review. The only change this PR will introduce is to upgrade the UI. The search index is already ready, as you can see from the deploy preview.

amarantha-k (approved):

PR #357, opened 2021-05-31
Add GTM script to header

amarantha-k:


amarantha-k:
@mojavelinux We added a Free Trial button to the header and need to add this script on page in order to track the event. Running this PR by you for a quick review.

mojavelinux:
The GTM script is already being loaded on the docs.couchbase.com site. By making this change, it will be added twice. Is there is reason for that?

See:

* https://github.com/couchbase/docs-ui/blob/master/src/partials/head-last.hbs#L10-L13
* https://github.com/couchbase/docs-ui/blob/master/src/partials/head-first.hbs#L6-L14

amarantha-k:
I had the same comment. The web team mentioned that the GTM script that is already being loaded only enables tracking and this additional change is needed to track specific events (button-click).

PR #385, opened 2021-07-27
update UI to navigate automatically on version select

mojavelinux:


mojavelinux:
This deploy preview proves that the change works, but was sent from the wrong branch. Closing and opening another to correct it.

PR #386, opened 2021-07-28
Update UI to navigate automatically on version change

mojavelinux:


mojavelinux:
@amarantha-k This is now ready to be merged so it will be in place when then production builds start back up again.

PR #391, opened 2021-08-10
add CSP

mojavelinux:


antora-playbook.yml
```diff
@@ -147,6 +147,7 @@ asciidoc:
   - ./lib/tabs-block.js
 ui:
   bundle:
-    url: https://github.com/couchbase/docs-ui/releases/download/prod-121/ui-bundle.zip
+    url: https://deploy-preview-116--cb-docs-ui.netlify.app/dist/ui-bundle.zip
```
osfameron, line 150: What does this entail? Looks like we'd be relying on a bundle published to netlify, rather than the one built in our repo?
Can you give some more details / pointers?

mojavelinux: Please do not review this PR. I'm testing an unreleased change to the UI and need to cross reference against the full site. That's why I have marked this as a Draft PR.

PR #394, opened 2021-09-02
3.2.1

RichardSmedley:


osfameron:
I didn't know about this nginx config, thought all the version update stuff happened at asciidoc/antora config level, would be good to run through this at some point :) 

But looks good as far as I can see 👍

maria-robobug (approved):
👍 LGTM, ditto on the `nginx` conf, wasn't even aware we were using it but makes sense!

PR #397, opened 2021-09-21
Update UI to v127 for language highlighting changes

maria-robobug:
Update the UI bundle to support language highlighting for the `nodejs` alias and the addition of `typescript`.


RichardSmedley (approved):
:+1:

PR #404, opened 2021-10-13
:page-alias: example

RichardSmedley:
DOC-9287


home/modules/contribute/pages/attributes-and-roles.adoc
```diff
@@ -113,6 +113,22 @@ Only requires for AsciiDoc files in the pages directory (not the partials direct
 
 page-aliases:: A comma-separated list of alternate page IDs that should redirect to this page.
 When you move a page, you might want to list the old page ID in this attribute so that the web server will redirect visitors from the old page to the new location.
```
simon-dew, line 115:
```suggestion
We recommend that, as far as possible, you avoid renaming or moving pages when editing the documentation.
If you do need to move or rename a page, you might want to list the old page ID in this attribute so that the web server will redirect visitors from the old page to the new location.
```

simon-dew (commented):
The expanded instructions LGTM. I'd like to suggest that we explicitly state that contributors should not move or rename files, if possible.

PR #419, opened 2021-10-30
upgrade UI to fix width of search results between 769px and 1024px

mojavelinux:
Sets the minimum width of the search results to 50vw so it doesn't shrink when the screen width is between 769px and 1024px.


RichardSmedley (approved):
LGTM :+1: 
Thanks @mojavelinux

PR #421, opened 2021-11-02
output log on failure in Netlify build

mojavelinux:


mojavelinux:
@maria-robobug This appears to work as we expect. If the Antora command fails for any reason, it will print out the contents of antora.log. Otherwise, antora.log will be published to the root of the site. Does that seem like a reasonable approach to you?

maria-robobug:
Hey @mojavelinux, yes I think that should be all we need for this 👍 .

maria-robobug (approved):

PR #423, opened 2021-11-03
Revert output log failure test.

maria-robobug:
Go too excited and forgot to amend the below PR after confirming that the test to make the build fail and output logs was successful!

https://github.com/couchbase/docs-site/pull/421/files


simon-dew (approved):
Looks like the current `idseparator` key so LGTM 👍

PR #424, opened 2021-11-03
[NO ISSUE] Build release/7.1 on staging site

simon-dew:


simon-dew:
Not sure why it hasn't built anything from `docs-analytics`...

simon-dew:
Okay, I don't know why it's not building anything from `docs-analytics`, but it works when I build it locally. I'm tempted to say this is ready for review and see if the problem persists on staging.

maria-robobug (approved):
Can't see where it says it can't build `docs-analytics`, seem good to me unless I'm missing something!

👍

simon-dew:
> Can't see where it says it can't build `docs-analytics`, seems good to me unless I'm missing something!

If you check out the deploy preview, go to **Server** > **Develop**, Analytics is not listed for any version

PR #427, opened 2021-11-04
Update to docs-ui bundle 135

osfameron:
This adds :page-toclevels: support.

### DIFF
--- a/antora-playbook.yml
+++ b/antora-playbook.yml
@@ -147,6 +147,6 @@ asciidoc:
   - ./lib/tabs-block.js
 ui:
   bundle:
-    url: https://github.com/couchbase/docs-ui/releases/download/prod-133/ui-bundle.zip
+    url: https://github.com/couchbase/docs-ui/releases/download/prod-135/ui-bundle.zip
 output:
   dir: ./public


maria-robobug:
Did we test this in staging? (staging branch)

RayOffiah (approved):

mojavelinux:
@maria-robobug As a word of advice, technically it's not necessary to test in staging as the PR already provides a deploy preview. As I understand, staging is supposed to be for testing content sources that are not yet ready for production, such as a new release.

osfameron:
@mojavelinux thanks - in this case we did want to check that the UI change didn't break anything.

Currently though the staging builds are broken, so we've mostly been focusing on that: e.g. https://app.netlify.com/sites/cb-docs-staging/deploys/61840706bde7a51ae757cf20
It looks as though only around 3% of the Antora log output is being shown which doesn't help...

mojavelinux:
> in this case we did want to check that the UI change didn't break anything.

But you already can. That's the whole point of the deploy preview. It gives you exactly the same thing you would get with staging. The only difference is that the URL is not branded.

mojavelinux:
>  It looks as though only around 3% of the Antora log output is being shown which doesn't help...

It's absolutely essential that the log messages be acted on. The messages are trying to communicate that there are problems with the docs, but no one seems to be addressing them...so the log continues to grow without bounds.

mojavelinux:
> Currently though the staging builds are broken

I think the staging builds are broken because a repository has been added that is causing Antora to run out of memory. I recommend reverting the content sources to what is in production, and adding them back in incrementally until you can figure out which repository is causing the problem. The Netlify deploy previews for the production environment work fine, so we can expect that the site will build successfully in Netlify.

osfameron:
@mojavelinux thanks:

* preview on netlify is perfect, yes that makes much more sense in this case as you say
* yes! the build logs is on my TODO after our recent team prioritisation, I was hoping to look at it today...

osfameron:
>  I recommend reverting the content sources to what is in production, then adding them back in incrementally until you can figure out which repository is causing the problem.

Good call thanks, will do this shortly

maria-robobug:
> As a word of advice, technically it's not necessary to test in staging as the PR already provides a deploy preview. As I understand, staging is supposed to be for testing content sources that are not yet ready for production, such as a new release.

Thanks @mojavelinux, I've clearly misunderstood the use of the `staging` environment in that case. If the Deploy Preview is enough to show validity of the UI changes then we're all good 👍 .

PR #428, opened 2021-11-05
tail antora log on failure

osfameron:
Cat seems to generate only ~920 lines of output, which is just 3%
of the 34,000 the local Netlify docker build image produces.

### DIFF
--- a/antora-playbook.yml
+++ b/antora-playbook.yml
@@ -47,21 +47,21 @@ content:
     start_path: docs
   - url: https://git@github.com/couchbase/docs-connectors-talend
   - url: https://git@github.com/couchbase/docs-analytics
-    branches: [release/7.1, release/7.0, release/6.6, release/6.5, release/6.0, release/5.5]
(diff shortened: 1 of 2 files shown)


mojavelinux:
I don't think it's cat that is the limit here. I think Netlify is cutting it off. Let's see.

We could just go back to not directing the log to a file. The reason we started doing it is because the messages were making the build log unwieldy. That was really a bandage.

osfameron:
> I don't think it's cat that is the limit here. I think Netlify is cutting it off. 
Agreed. Just as a hunch that if there's an error that's helpful it'll be towards the *end* of the file.

So this is another temporary bandage... to be reverted after it's helped/not helped debug this fail.

mojavelinux:
At this point, I think it might be best to take away the redirection and just let the messages go into the build log.

Once we're able to upgrade to Antora 3.0.0-beta.1, there are log controls built into Antora that we'll be able to use. So we can look forward to a better solution.

maria-robobug (approved):

PR #430, opened 2021-11-05
[NO ISSUE] build version 7.1 on staging site

simon-dew:
Second attempt. Removing all versions earlier than 6.0 from the staging site to cut down on build times, given that version 5.5 is now unsupported.

UPDATE: removing all versions earlier than 6.0 didn't work, but removing `dev-reorg` did. Seems we can't have `dev-reorg` and `release/7.1` at the same time.

### DIFF
--- a/antora-playbook.yml
+++ b/antora-playbook.yml
@@ -59,7 +59,7 @@ content:
     start_path: docs
   # NOTE docs-server is currently after other server repos so nav key wins
   - url: https://github.com/couchbase/docs-server
-    branches: [release/7.1, dev-reorg, release/7.0, release/6.6, release/6.5, release/6.0]
+    branches: [release/7.1, release/7.0, release/6.6, release/6.5, release/6.0]
   - url: https://github.com/couchbase/docs-sdk-common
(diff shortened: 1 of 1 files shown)


simon-dew:
After removing `dev-reorg` I experimented with reinstating the pre-6.0 versions. The build was successful in 23 minutes. To save processing time I'm minded to remove the pre-6.0 versions again on the staging build. Any thoughts?

RichardSmedley (approved):
:+1:

maria-robobug:
So just to understand, having removed the `dev-reorg` branch, could we not remove the pre-6.0 branches also to see if the build passes now?

Since `staging` seems to be used more to test new or experimental stuff maybe it should be a bit more lightweight?

RichardSmedley:
> So just to understand, having removed the `dev-reorg` branch, could we not remove the pre-6.0 branches also to see if the build passes now?

We can remove the pre-6.0 branches, as they'll be moving out of prod at the end of the month, so will be eventually consistent.

simon-dew:
Removed branches earlier than 6.0 again. Build now completes in 14 minutes. Shame about the dev-reorg branch, I'll just have to host that elsewhere.

PR #431, opened 2021-11-06
Rayoffiah/adding stem support

RayOffiah:
Adding support for Scientific, Technology, Engineering and Maths formula

### DIFF
--- a/antora-playbook.yml
+++ b/antora-playbook.yml
@@ -145,6 +145,7 @@ asciidoc:
   - ./lib/multirow-table-head-tree-processor.js
   - ./lib/swagger-ui-block-macro.js
   - ./lib/tabs-block.js
+  - '@djencks/asciidoctor-mathjax'
 ui:
   bundle:
     url: https://github.com/couchbase/docs-ui/releases/download/prod-132/ui-bundle.zip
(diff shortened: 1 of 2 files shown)


RayOffiah:
These checks failing are a little weird. Have I missed something?

maria-robobug:
Wonder if it would be useful to add the log output stuff in master as well - on staging it has been quite useful for debugging failures. Is it all working on your local build @RayOffiah.

RayOffiah:
> Wonder if it would be useful to add the log output stuff in master as well - on staging it has been quite useful for debugging failures. Is it all working on your local build @RayOffiah.

@maria-robobug 

Yes, that's the weird part. All seems to work fine.

mojavelinux:
Please do not use this package. That is not an official package of the Asciidoctor project and we cannot support that.

If you want STEM support, please follow the approach we have taken for the Asciidoctor docs. See https://github.com/asciidoctor/asciidoctor-docs-ui/blob/main/src/partials/footer-scripts.hbs#L3-L23 This allows you to enable stem support on a per page basis. It does link to the MathJax CDN, though it's possible to bundle MathJax like we do with other libraries.

mojavelinux:
To clarify, MathJax should be integrated into the UI and loaded at runtime. Trying to process STEM at build time (as this package does) is going to impact the build time substantially.

RayOffiah:
> To clarify, MathJax should be integrated into the UI and loaded at runtime. Trying to process STEM at build time (as this package does) is going to impact the build time substantially.

Understood.

I'll close the PR.

PR #434, opened 2021-11-15
Doc 9391 – moving release/5.5, release/5.1 and release/5.0 to archive master branch

RayOffiah:
Moved these to the archive site in a separate PR (433)

### DIFF
--- a/antora-playbook.yml
+++ b/antora-playbook.yml
@@ -50,7 +50,7 @@ content:
   - url: https://git@github.com/couchbase/docs-analytics
     branches: [release/7.0, release/6.6, release/6.5, release/6.0, release/5.5]
   - url: https://github.com/couchbase/couchbase-cli
-    branches: [cheshire-cat, mad-hatter, 6.5.x-docs, alice, vulcan, spock, 5.0.1]
+    branches: [cheshire-cat, mad-hatter, 6.5.x-docs, alice]
     start_path: docs
   - url: https://git@github.com/couchbase/backup
     branches: [cheshire-cat, mad-hatter, 6.5.x-docs, alice, vulcan, spock, 5.0.x]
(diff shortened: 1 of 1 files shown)


maria-robobug:
Hmm, I can still see the versions in the server docs dropdown in the `Deploy Preview`. Which is strange, not sure why that's happening?
Screen Shot 2021-11-16 at 9 52 37 AM

RayOffiah:
> Hmm, I can still see the versions in the server docs dropdown in the `Deploy Preview`. Which is strange, not sure why that's happening? Screen Shot 2021-11-16 at 9 52 37 AM

Right, just checking.

RayOffiah:
Okay, I get a different drop down.

I think the preview isn't actually doing anything with the version. If you have a look at the file `ui-model.yml` I think that is faking the menu.

maria-robobug (approved):
👍 LGTM now that I understand what's happening with the left hand nav server versions.
Not sure how we get around that as it can be quite misleading, but that's something to think about in future maybe.

RayOffiah:
The only real way to test it is to build the site locally using the real playbook. It's a bit of a trick; I'm trying it now to make sure it's not a genuine problem I've missed.

RichardSmedley:
The change should be made first to https://github.com/couchbase/docs-site/blob/staging/antora-playbook.yml
and that should show well whether or not there is a missing extra element.  ;)

RayOffiah:
Okay, staging! Got it.

maria-robobug:
See this is why I'm still a bit skeptical about `Deploy Preview` being enough to show things are working ok for production PRs. Especially if we are faking/mocking things in some places.

As @RichardSmedley mentioned, we've got staging to test things. I think we've even done this already there because we had issues with building too many branches, so Simon removed the old 5.x branches there already.

It seems to work fine there: https://docs-staging.couchbase.com/home/server.html

So this should be ok 🤞.

RayOffiah:
I've also removed the older versions of cloud-cli to match up to the server version.

maria-robobug (approved):

PR #435, opened 2021-11-17
reverting for broken .NET ref

RichardSmedley:


maria-robobug (approved):
Change LGTM, not sure why the Netlify build timed out though..

PR #441, opened 2021-11-26
Keep latest branches only and comment out unnecessary repos

maria-robobug:
Let's see how fast these builds can go! 🐎


osfameron:
built for speed!

PR #443, opened 2021-11-30
[DOC-9449] - Point to modules page in Node.js API doc

maria-robobug:
https://issues.couchbase.com/browse/DOC-9449

The new Nodejs 3.2 API doc front page seems quite confusing for users, let's point directly to the modules page where all the objects and functions are listed.


osfameron (approved):
👍

PR #446, opened 2021-12-08
Retrieve light icon set for fontawesome

RayOffiah:
Made a change to the doc site so that the light elements of Font awesome can be used.

### DIFF
--- a/scripts/package.json
+++ b/scripts/package.json
@@ -12,6 +12,7 @@
   },
   "optionalDependencies": {
     "@fortawesome/pro-solid-svg-icons": "~5.15",
-    "@fortawesome/pro-regular-svg-icons": "~5.15"
+    "@fortawesome/pro-regular-svg-icons": "~5.15",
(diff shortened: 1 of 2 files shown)


mojavelinux:
Exactly the right idea. Note that you should make a matching change to the UI so that these icons can be used in the UI preview. See https://github.com/couchbase/docs-ui/blob/master/gulp.d/tasks/build-preview-pages.js#L13-L29

RayOffiah:
> Exactly the right idea. Note that you should make a matching change to the UI so that these icons can be used in the UI preview. See https://github.com/couchbase/docs-ui/blob/master/gulp.d/tasks/build-preview-pages.js#L13-L29

Thanks, Dan.

I have the PR for that here:

https://github.com/couchbase/docs-ui/pull/128

But it's not allowing me to assign a reviewer 🙄. I'll @maria-robobug  to give it a look over.

Thanks for giving this the once over.

maria-robobug (approved):
👍

PR #453, opened 2021-12-14
Change developers.adoc and index-refresh.adoc into dot files

dgnicholson:
Both developer.adoc ([prod](https://docs.couchbase.com/home/developer.html)) and index-refresh.adoc ([prod](https://docs.couchbase.com/home/index-refresh.html)) appear to be defunct pages that are not consistent with the rest of the site and suffer from some styling and functional issues. These also rank in our top Google results, so cleaning them up will help with first time visitor experience. Making these dot files to ensure they are not required by someone else in the org before they're eventually deleted.


RichardSmedley (approved):
Thanks @dgnicholson :+1:

PR #459, opened 2022-01-06
Prepare CMOS for DP release

maria-robobug:
Setting this up here now, so that we are ready whenever the release goes ahead!


home/modules/ROOT/pages/cloud.adoc
```diff
@@ -77,6 +77,20 @@ The Service Broker implements the https://www.openservicebrokerapi.org/[Open Ser
 []
 xref:service-broker::index.adoc[Go to Couchbase Service Broker]
 
+ifdef::enable-cmos[]
```
maria-robobug, line 80: Don't think we actually use this page, but adding it in here as well anyway.

RichardSmedley (approved):
LGTM

PR #471, opened 2022-01-19
Update Jenkinsfile

RayOffiah:
Removing the redirection so we can see the errors in the Jenkins console.


RayOffiah:
Okay, it looks like the checks are failing, possibly for the same reason that the build is failing? --> @maria-robobug

maria-robobug (approved):

PR #474, opened 2022-01-21
Show deploy preview errors on failure

maria-robobug:
Specifically doing this on failure because the Deploy Previews are done on Netlify and we don't want to fill up the logs with all our warnings every time a build runs.


RayOffiah (approved):
👍🏾

PR #476, opened 2022-01-24
RayOffiah-patch-adding-stacktrace-option

RayOffiah:
Added the stacktrace option to see if we can get a few more details out of the failing checks.


RayOffiah:
No longer needed

PR #493, opened 2022-02-03
Update Jenkinsfile

RayOffiah:
Installing asciidoctor-kroki before the antora is run


maria-robobug:
Guess no point waiting on Netlify because this is Jenkins specific, so we won't know if it works until it's merged 😅 🤞 !!

maria-robobug (approved):

PR #500, opened 2022-02-07
Update UI production to 143

RayOffiah:
Fixes problem with stem CSS spacing on Firefox


ceejatec:
@RayOffiah Not sure why you requested a review from "cb-robot"? Was that a mistake? It notified me, but I don't think I can review this...

PR #503, opened 2022-02-11
upgrade to Antora 3.0.1

mojavelinux:
- upgrade to Antora 3.0.1
- switch from using custom site generator to Antora extensions
- convert script to remove private content sources to an extension


lib/exclude-private-content-sources-extension.js
```diff
@@ -0,0 +1,12 @@
+'use strict'
+
+module.exports.register = function () {
+  if (this.getVariables().playbook.env.GIT_CREDENTIALS) return
+  this.once('playbookBuilt', function ({ playbook }) {
```
RayOffiah, line 5: This is very cool 👍🏾.

I'm seeing ways that the site content can be used to build all sorts of other stuff.

RayOffiah (approved):

lib/exclude-private-content-sources-extension.js
mojavelinux: Indeed. I think a next step is to integrate the icon loading directly into Antora so it doesn't have to be done as a separate script. But I think it's best to make that change in a separate PR.

maria-robobug (approved):
👍 LGTM, didn't see any glaring issues on the preview site from my quick look so far!

mojavelinux:
I've also updated the README. Let me know if you're ready to proceed and I'll merge. I'll have to monitor the production build since there is currently no way to test that without pushing the changes to master. But since we are only changing how Antora is run, the worst that can happen is that the build stops early and I'll need apply a fix.

mojavelinux:
We're really struggling to fit the Netlify build in within the time limit.

maria-robobug:
> We're really struggling to fit the Netlify build in within the time limit.

Yeah still something we haven't managed to address, guessing this happens because of all the warnings and errors... 😭 
I think the limit is like 15mins right?

RayOffiah:
But it does seem to carry on building it though.

maria-robobug:
@mojavelinux, happy for this to be merged, when you're ready 👍 .

mojavelinux:
I'll proceed!

mojavelinux:
Production is up and running on Antora 3.0.1!

mojavelinux:
I'll follow-up with a PR to make the same change in staging.

mojavelinux:
I have applied the update to staging.

I also have good news. The upgrade reduced the build time in production from 27 minutes (https://docs-build.sc.couchbase.com/job/Antora/job/docs.couchbase.com/job/master/1799/) to 19 minutes (https://docs-build.sc.couchbase.com/job/Antora/job/docs.couchbase.com/job/master/1802/). Some of that depends on how active the Jenkins server is at the time, though I expect that improvement to hold.

RayOffiah:
That's great work, Dan. Thanks!

maria-robobug:
🥳 Fantastic news, thanks @mojavelinux!

PR #507, opened 2022-02-15
DOC-9701-staging Add new global attribute to represent SQL++

RayOffiah:
Added new attribute

### DIFF
--- a/antora-playbook.yml
+++ b/antora-playbook.yml
@@ -157,6 +157,7 @@ asciidoc:
     xrefstyle: short
     enterprise: https://www.couchbase.com/products/editions[ENTERPRISE EDITION]
     community: https://www.couchbase.com/products/editions[COMMUNITY EDITION]
+    sqlpp: SQL++
     kroki-server-url: http://3.91.133.254:9500
     kroki-fetch-diagram: true
   extensions:


maria-robobug:
Not sure if it's due to this change from Dan, https://github.com/couchbase/docs-site/commit/a7be5a36485d400a6e4bd6010f879c9b6b9d8bb8#diff-0ede03e58603c03c400bde34263582dbb4a75248e7d1f52cdee1729c44398cabL6-L11

But let's see...

maria-robobug:
Oooh, I can see logs now, maybe we'll see what's wrong if it fails again: https://app.netlify.com/sites/cb-docs-staging/deploys/620be3f1be5e403a4d105b38#L80
👍

maria-robobug:
```
6:03:37 PM: Build exceeded maximum allowed runtime 
```
Well antora built things fine I think. Just seems Netlify timeout is going to be an even bigger problem...I'll merge 👍 .

RayOffiah:
Okay, thanks for that.

I was wondering it it's attempts to run the Kroki stuff is causing a problem, since it can't reach the server. That won't help with the timeouts.

maria-robobug:
> I was wondering it it's attempts to run the Kroki stuff is causing a problem, since it can't reach the server. That won't help with the timeouts.

It's quite possible ya know, didn't think of that... 🤔 . 
Let's keep an eye on it.

PR #509, opened 2022-02-19
Versions update

RichardSmedley:
Maybe take 7.1 out of Server.


maria-robobug (approved):
Seems fine to keep 7.1 for Server, since this is for local use anyway?
Unless we are referencing this as an example anywhere visible to users...

RichardSmedley:
> Seems fine to keep 7.1 for Server, since this is for local use anyway? Unless we are referencing this as an example anywhere visible to users...

It's on this page - https://docs.couchbase.com/home/contribute/test-site.html#lightweight-local-playbook

PR #517, opened 2022-02-25
Ray offiah add kroki support files

RayOffiah:
These don't actually do anything. I just thought it would be a good idea to keep the files used to build the service.


maria-robobug (approved):
Ah very nice! 
Yeah good to know we can play around with it in Docker if needed too 👍

PR #527, opened 2022-03-24
[NO ISSUE] Revert attribute for query language

simon-dew:
We should not change the name of the query language until (a) we have introduced this attribute consistently across the entire documentation set, as far as possible; and (b) the plan to change the name has actually been finalized.


simon-dew:
Good points, but we still need to set the attribute back to N1QL, so I'm going merge this for now.

PR #531, opened 2022-04-11
Update failure routine

maria-robobug:
As per Dan's suggestion I'm updating the failure routine logic to help clear some space on Jenkins...


mojavelinux:
Nice work!

It's possible that the failure clause was always incorrect. If that's the case, I apologize retroactively for having made that error and not testing it. That said, I maintain other Jenkins builds that are able to use `deleteDir()` without the `dir` enclosure, so it seems to have been correct at some point. What we can be sure of is that the updated clause is a guarantee ;) Onward!

PR #534, opened 2022-04-11
Revert delete directory attempts

maria-robobug:
Doesn't seem to be working, so at this point I'm going to revert to what was there initially.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -135,9 +135,7 @@ pipeline {
       githubNotify credentialsId: githubApiCredentialsId, account: githubAccount, repo: githubRepo, sha: env.GIT_COMMIT, context: 'continuous-integration/jenkins/push', description: 'The Jenkins CI build succeeded', status: 'SUCCESS'
     }
     failure {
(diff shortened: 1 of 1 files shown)


osfameron (approved):
Retreeeeeeeat!

mojavelinux:
I'm pretty confident that what was there before won't work as it can't run `deleteDir` (anymore?) without an enclosure. However, it's unclear from the Jenkins docs what the right enclosure is.

...but it would be better if we left the workspace in tact for now so we can understand how to reduce its size without remove it entirely. So the fact that deleteDir is not working right now is actually a *good thing*&trade;.

mojavelinux:
What we've learned is that the content cache in .cache/antora/content was 15G. After running `git gc` on each repository in that directory, the total size reduced to 1.4G. So we're going to need to do some `git gc` housekeeping periodically to keep the size of the cache under control.

maria-robobug:
@mojavelinux fantastic, so we should use that instead of the `deleteDir()` approach then?
Happy to add this groovy script for now, but the Antora extension approach would be nice too!

PR #546, opened 2022-04-29
Nodejs 4.1

RichardSmedley:


etc/nginx/snippets/rewrites.conf
```diff
@@ -6,7 +6,7 @@ set $current_version_sdk_go_api '2.5.0';
 set $current_version_sdk_java_api '3.2.7';
 set $current_version_sdk_jvm_core_api '2.3.0';
 set $current_version_kotlin_client_api '1.0.0-dp.7';
-set $current_version_sdk_nodejs_api '4.0.0';
+set $current_version_sdk_nodejs_api '4.1.0';
```
maria-robobug, line 9: Guess we need to also revert the dotnet redirect?
https://github.com/couchbase/docs-site/pull/546/files#diff-db7cb9ef457c55b2c042d8af81f4b0db307047377cbb7a9ca7461e55f31b0642R4

RichardSmedley: I don't think it will affect many people in practice.

maria-robobug (approved):
👍

PR #552, opened 2022-05-07
Inline jira:: macro

osfameron:
Instead of having to write something like the following in release
notes:

 * https://issues.couchbase.com/browse/PYCBC-849[PYCBC-849]:

Now we can use a macro:

 * jira::PYCBC-849[]

Full usage details and configuration in the supplied `extensions.adoc`
(not added to nav for now, as it's internal. Is there a better place
to put this?)

NB: the configuration is currently done inside the asciidoc extension,
as these don't accept config.
A better solution might be to convert this to an *Antora* extension
(which does accept config) which simply reads the config, and registers
a properly configured asciidoc extension dynamically?


RayOffiah:
I like it! Great idea, especially for the Release Notes page.

It feels like it should be an Antora extension though, at least as some sort of wrapper so that the config can go in the playbook. I'm not too familiar with Antora extensions yet, so I'm not sure. I would probably hold off until we get a steer from @mojavelinux. 

But this would be really useful, definitely.

Oh, and is it possible to have something like 

`jira::AV-8828[this Jira]`

RayOffiah:
But this is a really good general purpose extension, so I would stick to making it a general purpose AsciiDoc module with an Antora wrapper (if possible) to get the params from the playbook.

osfameron:
> jira::AV-8828[this Jira]
@RayOffiah yeah, I wondered about that... not sure if it should *replace* the default text (the ticket number) or end up with something like "this Jira (AV-8828)" ?

Antora wrapper sounds good, I have *an idea*, not sure if it'll work, will timebox 30 mins, and then over to Dan for more guidance :)

RayOffiah:
I was thinking that if you had:

`jira::AV-8828[]`

The you'd just end up with the ticket number. But if they put text in the brackets, then you should just put whatever is in the brackets as the link text, even if they don't include the actual ticket number in the text.

Yup, 30mins sounds reasonable :-)

osfameron:
@RayOffiah OK, the `[]` change should be simple, I can do that!

RayOffiah:
LGTM 👍🏾, but probably wait for Dan to take a look.

RayOffiah (approved):

home/modules/contribute/pages/extensions.adoc
```diff
@@ -0,0 +1,39 @@
+= Site Extensions
+
+Test and document your Asciidoctor.js or Antora extensions here.
+
+== inline JIRA macro
```
maria-robobug, line 5:
```suggestion
== Inline JIRA macro
```

home/modules/contribute/pages/extensions.adoc
```diff
@@ -0,0 +1,39 @@
...
+[source,asciidoc]
+----
+== Fixed issues:
+
+* jira:JSCBC-1234[]: reversed the polarity of the neutron flow
+* jira:AV-2345[]: frobnicated the gostaks
+* jira:DOC-999[The dreaded DOC-999 ticket...]: implemented jira macro
+----
+
+[,yml]
```
maria-robobug, line 18:
```suggestion
[source,yml]
```

maria-robobug (approved):
👍 LGTM, some minor comments. 
We should add this change in the staging branch too.

home/modules/contribute/pages/extensions.adoc
mojavelinux: FYI: The `source` style is now inferred if a language is specified.

osfameron: Good to know - though I'm happy with keeping it explicit personally, let's discuss within the team.

PR #572, opened 2022-06-14
Asciidoc markdown block

osfameron:
Block extension to filter text like:

    [markdown]
    --
    Filter some text with [Markdown](https://commonmark.org/help/) syntax.
    --

This is not implemented with a full Markdown parser.
See https://github.com/asciidoctor/kramdown-asciidoc/issues/7
with a link to the "naive series of regexes" used as starting point.

(And note that we use Open structural context
https://docs.asciidoctor.org/asciidoc/latest/blocks/delimited/#summary-of-structural-containers
with `--` delimiters, and headings don’t work inside these.)

This feature is intended for handling OpenAPI specs, which can contain Markdown,
however openapi-generator has poor Asciidoc handling
https://github.com/OpenAPITools/openapi-generator/issues/11396
so instead we add the block delimiters in the template, and let the block filter handle it.

### DIFF
--- a/antora-playbook.yml
+++ b/antora-playbook.yml
@@ -171,6 +171,7 @@ asciidoc:
   - ./lib/multirow-table-head-tree-processor.js
   - ./lib/swagger-ui-block-macro.js
   - ./lib/tabs-block.js
+  - ./lib/markdown-block.js
   - asciidoctor-kroki
 ui:
   bundle:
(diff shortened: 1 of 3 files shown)


RayOffiah (approved):
Hah! Wow! 

Okay, the module registration looks good to me, and the mdtoac.js test seems to do the business, so I'm good with it.

PR #585, opened 2022-07-18
Adding extension to enable callouts without associated listing block (for staging).

RayOffiah:
https://github.com/asciidoctor/asciidoctor/issues/3359

### DIFF
--- a/lib/asciidoctor-callout-list-block.js
+++ b/lib/asciidoctor-callout-list-block.js
@@ -1,0 +1,18 @@
+module.exports = function (registry) {
+
+    registry.treeProcessor(function () {
+
+        const self = this
+
+        self.process(function (document) {
+
+            document.findBy({'context': 'olist', 'style': 'calloutlist'}, function (list) {
+
+                list.context = list.node_name = 'colist'
+
+            })
+            return document
+        })
+
+    })
+}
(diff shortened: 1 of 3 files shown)


lib/asciidoctor-callout-list-block.js
```diff
@@ -0,0 +1,18 @@
+module.exports = function (registry) {
+
+    registry.treeProcessor(function () {
+
+        const self = this
+
+        self.process(function (document) {
+
+            document.findBy({'context': 'olist', 'style': 'calloutlist'}, function (list) {
```
maria-robobug, line 9: Is `olist` meant to be `colist`?

maria-robobug (approved):
👍 LGTM, just a minor question.

lib/asciidoctor-callout-list-block.js
RayOffiah: Good question.

What I'm doing is taking an ordered list, and converting into a `colist`, so this line is doing a search for any ordered list in the document which has a `calloutlist` style associated with it.

If it finds one then it converts the ordered list into a `colist` which the Asciidoctor parser will see and then add a load of extra tables and styles to make it into a proper `colist`.

So the `olist` is correct because that's what I'm searching for  :-)

RayOffiah:
To use it:

```asciidoc
[calloutlist]
. First line.
. Second line.

```

PR #586, opened 2022-07-27
AV-39681 stage July 2022 changes for capella

osfameron:


RayOffiah (approved):
Yup, that should do it.

osfameron:
Tweaked to remove main from the staging build, left the original in a yaml comment for reference.

RayOffiah (approved):

PR #591, opened 2022-08-02
rayoffiah-update-staging-to-151

RayOffiah:
Also upgrading the callout extension; the new. version allows for the creation of callout lists without a listing block attached.


maria-robobug (approved):
👍 LGTM

PR #599, opened 2022-08-31
N1QL ---> SQL++

maria-robobug:
Changing this to SQL++, not sure if we've started using it, but I have in SDK...


osfameron (approved):
This looks fine - though will depend on how many places we already use it in our codebase?

(I searched in my locally checked out repos - and could only find it for docs-couchbase-lite, though it's defined locally there already, so that won't affect it).

maria-robobug:
Yeah, that's why I added everyone to the PR - I'll merge this, worst case I can just revert the change if anyone is opposed to this.

PR #601, opened 2022-09-05
DOC-10357 CAO: Update SG docs to use bootstrap config

RayOffiah:
Change to staging to check change.


maria-robobug (approved):
🤞 Let's see how it goes!

PR #613, opened 2022-09-23
Update the ui bundle to prod-154

maria-robobug:
https://github.com/couchbase/docs-ui/commit/d9467a167ec308c85d90f037045d755a087a92d4

This updates the UI bundle to support the page type tags change. Starting with staging so we can test it out.


sarahlwelton (approved):
Let's do it!

PR #616, opened 2022-09-29
update-staging-to-ui-157

RayOffiah:
This adds brings back the SQL++ language to the syntax highlighter, using sqlpp as the language name.


maria-robobug (approved):
👍

PR #617, opened 2022-09-29
Update prod to UI 157

RayOffiah:
This adds brings back the SQL++ language to the syntax highlighter, using sqlpp as the language name.


maria-robobug (approved):

osfameron (commented):
Looks like you only need the last commit in there?
Will all the merges/reverts be cleaned up before merging?

RayOffiah:
> Looks like you only need the last commit in there?
> Will all the merges/reverts be cleaned up before merging?

Could do.

What's the preferred way to do that?

osfameron:
> What's the preferred way to do that?

Ah, well "Squash and merge" should be fine for the actual merge!

But I tend to bang on about Git commit history (sorry) cos I got used to it as a dev and it makes things easier to review (also cos often if you see a whole load of unrelated commits, it's a warning flag that you might be trying to merge into the wrong branch! So I really like to check.)

I'd rebase instead of merging as I go, keeps everything cleaner.
It's the proposal for our standard workflow, discuss at https://docs.google.com/document/d/14IvKRgcjtw2RSjer86rfNuLuHDqIXcZvCWVAb4j3900/edit# !

(It's easy enough to fix afterwards, but harder to explain. Happy to chat on a Zoom if you're interested though)

RayOffiah:
RIght, so we can merge this and fix it afterwards?

maria-robobug:
^ You can always use "Squash and merge" to clean things up, so that should be fine.

But I also would suggest getting into the habit of rebasing, to keep commit history sane etc 👍 .

RayOffiah:
👍🏾

PR #620, opened 2022-10-11
Add icon script back in steps and update fontawesome token variable

maria-robobug:
It's possible that when IT recently sorted the Jenkins disk space issues, the information needed for fontawesome was somehow lost. https://couchbase-internal.zendesk.com/hc/en-us/requests/11756

Ray has added the `FONTAWESOME_NPM_AUTH_TOKEN` env variable on the Jenkins box (via ssh) in ~/.profile, so that should be permanently available for Jenkins to use for fontawesome.

This has been tested on `staging` and the icons are showing successfully, so I'm making this change in prod now.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -82,7 +82,7 @@ pipeline {
           }
         }
         withCredentials([fontawesomeNpmTokenCredentials]) {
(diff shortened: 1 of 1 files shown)


RayOffiah (approved):
LGTM 👍🏾

PR #622, opened 2022-10-14
Uncomment icon script step

maria-robobug:
Now that Ray has obtained a new fontawesome token for us, this should work.


RayOffiah (approved):

RayOffiah:
I approve!

PR #640, opened 2023-01-02
Support for multiple macros on the same page

LukeCz:
Hi, the current design assumes there is only one hardcoded swagger-ui div per page. When the macro is used multiple times, the content goes always to the 1st div. This proposed change is to parametrize the div name and script name using a simple global counter.


osfameron (commented):
Hi @LukeCz, this is interesting thank you for the contribution.

I don't know of cases where we want to have multiple SwaggerUI macros, does this PR respond to a particular requirement or bug on a specific page in the Couchbase docs?

If it's a general enhancement for all users of Antora, it might be worth suggesting to https://gitlab.com/antora/antora/ (and we do periodically update our extensions from this upstream as required).

PR #643, opened 2023-01-07
AV-50499 embargo pages mockup

osfameron:
This requires:

* antora 3.1.1 or some release after https://gitlab.com/antora/antora/-/commit/36ae21fae5654b4f9443e6a5cc0a48006c9ec4d7

* looks like we're not ready to upgrade to that (notice how I've had to comment out source-url-include-processor.js to test this) and have pinged Opendevise https://support.opendevise.com/support/tickets/170

### DIFF
--- a/home/modules/contribute/pages/extensions-embargoed.adoc
+++ b/home/modules/contribute/pages/extensions-embargoed.adoc
@@ -1,0 +1,5 @@
+:page-embargo: EMBARGOED
+= This page is embargoed!
+
+This page should not be output.
+See `lib/embargo.js` and AV-50499
--- a/home/modules/contribute/pages/extensions-embargoed.adoc
+++ b/home/modules/contribute/pages/extensions-embargoed.adoc
@@ -1,5 +1,9 @@
+ifndef::flag-test-embargo[]
 :page-embargo: EMBARGOED
+endif::flag-test-embargo[]
+
 = This page is embargoed!
 
 This page should not be output.
+(Because the embargo attribute is controlled by an `ifndef::flag-test-embargo[]` and the attribute is not supplied).
 See `lib/embargo.js` and AV-50499
--- a/antora-playbook.yml
+++ b/antora-playbook.yml
@@ -4,6 +4,8 @@
 antora:
   extensions:
   - '@antora/site-generator-ms'
+  - ./lib/embargo.js
+
 site:
   title: Couchbase Docs
   url: https://docs.couchbase.com
--- a/home/modules/contribute/pages/extensions-embargoed-ok.adoc
+++ b/home/modules/contribute/pages/extensions-embargoed-ok.adoc
@@ -1,0 +1,8 @@
+ifdef::flag-test-embargo[]
+:page-embargo: EMBARGOED
+endif::flag-test-embargo[]
+
+= This page should not be embargoed!
+
+This page should be output, because the embargo attribute is controlled by an `ifdef::flag-test-embargo[]`
+See `lib/embargo.js` and AV-50499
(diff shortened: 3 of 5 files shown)


osfameron:
@mojavelinux this is the PR for page embargo, suggestions for improvement welcome (assuming you're able to see this repo?)

mojavelinux:
There's actually a simpler way to do this. All you need to do is delete the `out` property from the virtual file. Antora will not publish files that do not have the `out` property. It's best to delete the `pub` property as well.

> By removing the out property from the page, it prevents the page from being published, but is still referencable using an include directive. Alternately, you could choose to remove the page entirely from the content catalog using removeFile.

However, applying this extension during navigationBuilt is too late to prevent xrefs to that page. If you want to do that, you need to listen for the contentClassified event. But then you will need to look for the attribute entry in the raw AsciiDoc source since it has not yet been parsed.

osfameron:
Thanks @mojavelinux - I don't quite understand the simpler way: can I avoid using the extension at all and remove the attributes in Asciidoc markup?

Otherwise, it looks like I'd simply replace:

  `contentCatalog.removeFile(page)`

with

  `delete page.pub; delete page.out` ?

(The advantage presumably being that we don't need to wait for 3.1.x to do this?)

> But then you will need to look for the attribute entry in the raw AsciiDoc source since it has not yet been parsed.
I don't want to do that, as the embargo would be in an ifdef (e.g. feature flag) so would require fully parsing the raw source.
(So the tradeoff of possible dangling xrefs is an acceptable minor risk I think)

mojavelinux:
I was merely pointing out that you can avoid `removeFile` by removing the `out` property on the file instead.

> I don't want to do that, as the embargo would be in an ifdef (e.g. feature flag) so would require fully parsing the raw source. (So the tradeoff of possible dangling xrefs is an acceptable minor risk I think)

:+1:

osfameron:
@simon-dew @sarahlwelton Please review if the embargo feature looks useful for your 🦓 work.
I'll test against Staging first before merging.

home/modules/contribute/pages/extensions-embargoed.adoc
```diff
@@ -0,0 +1,5 @@
+:page-embargo: EMBARGOED
```
simon-dew, line 1: Can this page attribute line be wrapped within `ifdef` / `ifndef` directives?

(I'm sure it can, just checking)

simon-dew (commented):
Just one comment. (I can't speak to the content of the JavaScript.)

home/modules/contribute/pages/extensions-embargoed.adoc
osfameron: Yes! But you're right, that'd be a much better and more reassuring test... 👍 
I'll add another example page shortly...

osfameron:
Hmm, our choices for reviewing Javascript are:

- trust me/previous review by Dan/staging build
- review it
- ask Dan to review it
- find someone in engineering to review it

I think we want to be more self-sufficient in Docs team if we can.
@sarahlwelton @RichardSmedley @simon-dew do any of you feel competent to have a look, possibly on a Zoom call together?

sarahlwelton (approved):
> Hmm, our choices for reviewing Javascript are:
> 
> * trust me/previous review by Dan/staging build
> * review it
> * ask Dan to review it
> * find someone in engineering to review it
> 
> I think we want to be more self-sufficient in Docs team if we can. @sarahlwelton @RichardSmedley @simon-dew @tonyjhillman do any of you feel competent to have a look, possibly on a Zoom call together?

So, I know enough JS to be able to follow what's going on with the code, and I really don't see anything glaringly wrong here. We already got a good suggestion to prevent the publishing from Dan Allen, and the logic here seems sound to me - iterate through each page once we've got the nav built and look for the page-embargo attribute to contain content, then remove the attributes that publish the page if it does. 

@osfameron and I discussed whether there might be a better event to use than navigationBuilt, but that might be something to dig into later, as we can take steps to mitigate the risk it causes through writer education and just using the ifdefs appropriately for links. 

👍 from me.

PR #645, opened 2023-01-09
Updating ZD creds

RichardSmedley:
DOC-10718


osfameron (approved):
Note that the new credentials are 1 character shorter - please check that's correct, but otherwise, looks fine.

PR #647, opened 2023-01-16
Spark 3.2 (3.3 in main)

RichardSmedley:


osfameron (approved):
👍 to unblock you as looks fine, but I don't understand what "3.3 in main" means.
(Ideally would be commented in commit description)

RichardSmedley:
> +1 to unblock you as looks fine, but I don't understand what "3.3 in main" means. (Ideally would be commented in commit description)

Yes, probably should have. 
It's https://github.com/couchbase/couchbase-spark-connector/blob/master/docs/antora.yml#L3

PR #653, opened 2023-01-27
DOC-10775: Docs infrastructure for 7.2.0

simon-dew:
* Add separate branches for 7.2 prerelease and 7.1
* Remove branches for 6.0

Docs issue: [DOC-10775](https://issues.couchbase.com/browse/DOC-10775)

### DIFF
--- a/antora-playbook.yml
+++ b/antora-playbook.yml
@@ -62,19 +62,19 @@ content:
     start_path: docs
   - url: https://git@github.com/couchbase/docs-connectors-talend
   - url: https://git@github.com/couchbase/docs-analytics
-    branches: [release/7.1, release/7.0, release/6.6, release/6.5, release/6.0]
+    branches: [release/7.2, release/7.1, release/7.0, release/6.6, release/6.5]
   - url: https://github.com/couchbase/couchbase-cli
-    branches: [master, cheshire-cat, mad-hatter, 6.5.x-docs, alice]
+    branches: [neo, 7.1.x-docs, cheshire-cat, mad-hatter, 6.5.x-docs]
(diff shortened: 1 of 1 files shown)


simon-dew:
@RichardSmedley you might want to have a look at the line for `docs-sdk-common`. Is there a branch specifically for release 7.2?

RichardSmedley:
> @RichardSmedley you might want to have a look at the line for `docs-sdk-common`. Is there a branch specifically for release 7.2?

@simon-dew  sdk-common is branch-named after the Server version that's roughly equivalent to the SDK API version.
SDK 3.4 lines up with 7.1.2 (the first time we've had to use a patch version).
SDK 3.5 will be a later-than-7.2 version, aiui,
...so no need for an sdk-common 7.2.

osfameron (approved):
Approving to unblock.
For cli and backup, where we've removed `master`, that means the preceding `master` has been removed to 7.1.x-docs, and you've persuaded the engineering teams to work in `neo` instead of a generic master?

simon-dew:
> For cli and backup, where we've removed `master`, that means the preceding `master` has been removed to 7.1.x-docs, and you've persuaded the engineering teams to work in `neo` instead of a generic master?

It means that the focus of development on the main branch has moved on to 7.5 or later. I'm not even tracking that in the docs staging build. 

Release 7.1 and 7.2 are both built from the `neo` branch. For the sake of documentation, I've persuaded the engineering teams to create a separate `7.1.x-docs` branch for the release 7.1 docs, and we will use the `neo` branch for the release 7.2 docs.

Thanks for the 👍

PR #662, opened 2023-03-23
Update antora-playbook.yml to use new UI release

sarahlwelton:
Now that I've come back to this, just want to update the UI bundle to change the [abstract] styling.


RichardSmedley (approved):
:+1:

PR #666, opened 2023-03-31
Update capella layout

osfameron:
This change points the prod playbook to the new layout introduced by https://github.com/couchbasecloud/couchbase-cloud/pull/18906

As we only have a single `start_path`, the user-visible site will be identical to before, as can be tested on docs-staging.


RichardSmedley (approved):
LGTM (once the Cloud repo change gets merged).  :+1:

osfameron:
This change is no longer needed

PR #667, opened 2023-04-04
Add capella and elixir branches to antora-playbook.yml for Staging

sarahlwelton:


workspace/search/search-capella
osfameron, line 1: did you mean to add these workspaces as submodules?

osfameron (changes requested):
Oh, yeah, this is docs-site, we don't want those workspace submodules

workspace/search/search-capella
sarahlwelton: Oh whoops let me fix that.

simon-dew (changes requested):
Just one addition please.

(I need to keep the Data API reference material in cb-swagger for the moment, which means it needs to have capella and elixir branches for now &mdash; which means it needs to move here. This may be just a temporary solution.)

.gitignore
```diff
@@ -13,5 +13,6 @@ xref-validator.log
 **/package-lock.json
 **/lab.sh
 **/.DS_Store
+/workspace
```
osfameron, line 16: 👍

osfameron (approved):
thanks 👍

simon-dew (approved):

PR #669, opened 2023-04-18
tutorial link test

RichardSmedley:


sarahlwelton (approved):
Looks good, as discussed on Slack, If it works in local builds, let's go!

antora-playbook.yml
```diff
@@ -22,7 +22,7 @@ site:
         { "title": "CMOS", "components": ["cmos"] },
         { "title": "Service Broker", "components": ["service-broker"] },
         { "title": "SDKs", "startPage": "home::sdk.adoc", "components": ["*-sdk", "cxx-txns", "elasticsearch-connector", "kafka-connector", "spark-connector", "tableau-connector"] },
-        { "title": "Tutorials", "startPage": "tutorials::index.adoc", "components": ["tutorials"] }
+        { "title": "Tutorials", "url": "https://developer.couchbase.com/tutorials/", "components": ["tutorials"] }
```
osfameron, line 25: Just to not that I'm not sure if the `components` key is needed, for my test I kept the key but set the value to `[]`.
This seems fine, and worth a test on staging 👍

osfameron (approved):

PR #671, opened 2023-04-24
Template block macro (handlebars)

osfameron:
See https://docs-staging.couchbase.com/home/contribute/extensions-template.html for a built demo of this.
(@mojavelinux I'll ping you the password for this separately)

This extension enables a `[template]` block which formats a specified .json file against a .hbs Handlebars template.

```
[template,example$some-data.json]
--
include::partial$templates/test.hbs[]

or, e.g:

{{#each items}}
* {{name}}
{{/each}}
--
```

See https://issues.couchbase.com/browse/DOC-10720 for a motivating example (metrics).

Other use-cases keep coming up:

 * OpenAPI to configuration schema generation, (currently done dynamically at e.g. https://docs.couchbase.com/sync-gateway/current/configuration-schema-database.html#_database )

 * Jira report -> release notes generation

etc.
    `


malarky (approved):
Hi @osfameron, as discussed, this looks like it definitely fulfils the primary requirement, which was to replace the current metrics reference tables with the JSON delivered from the builds.  It also gives us a good starting point to be able to improve the docs with (for example) improved styling, version tags, deprecation notices, etc.

home/modules/contribute/examples/kv-metrics.json
```diff
@@ -0,0 +1,32 @@
...
+        "type": "gauge",
+        "unit": "seconds"
+    },
+    "kv_bg_wait_seconds": {
+        "help": "bg fetches waiting in the dispatcher queue",
+        "type": "gauge",
+        "unit": "seconds"
+    },
+    "kv_cmd_duration_seconds": {
+        "help": "",
```
leeming, line 18: (minor) Missing help for this example

home/modules/contribute/pages/extensions-template.adoc
```diff
@@ -0,0 +1,44 @@
+= Demo of template extension
```
leeming, line 1: (Minor) Niggle around the name "template", but that may just be the lack of familiarity with the tooling. Seems an overloaded term

lib/template-block.js
```diff
@@ -0,0 +1,32 @@
+module.exports.register = function (registry, config) {
+const handlebars = require("handlebars")
```
leeming, line 2: (very minor) incorrect indentation

leeming (approved):
Looks good overall as it seems to hit the functional ask. Some minor points, but not worth holding up progress for them, more FYI/opinion.

lib/template-block.js
```diff
@@ -0,0 +1,32 @@
...
+  
+// [template,example$foo.json]
+
+  registry.block(function () {
+    var self = this
+    
+    self.named('template')
+    self.positionalAttributes('resourceId')
+    self.onContext(['listing', 'open', 'paragraph'])
```
mojavelinux, line 11: I would recommend only using the open block (and maybe the paragraph). That's because the output becomes an open block. If it were used on a listing as is, you'd be changing the content model, which is not friendly to tooling. This could be resolved by detecting the original content module (by looking for `cloaked-context`), but it would be best to make that a separate enhancement.

lib/template-block.js
```diff
@@ -0,0 +1,32 @@
...
+        var data = JSON.parse(target.contents)
+        var source = reader.getLines().join('\n')
+        
+        var template = handlebars.compile(source)
+        
+        var output = template(data)
+        return self.createOpenBlock(parent, output)
+      }
+      else {
+        return self.createOpenBlock(parent, `Data file ${attributes.resourceId} not resolved`)
```
mojavelinux, line 27: It would be best to log an message here. Refer to these pages to learn how:

* https://docs.asciidoctor.org/asciidoctor/latest/extensions/logging/
* https://docs.asciidoctor.org/asciidoctor.js/latest/processor/logging-api/

mojavelinux:
In general, this seems sound to me. I left some comments in addition to the ones by Andrew. If those could be addressed, I'd give this a :+1:.

home/modules/contribute/examples/kv-metrics.json
osfameron: That's in the original data, so left as-is to test 👍 
I might remove the field entirely, to check that the templating handles missing fields gracefully...

lib/template-block.js
osfameron: That's fine, I've previously used only `open` (for the "Markdown-ish" block) but I think I copied this from a doc/example in case it was best practice... will revert to just open 👍

lib/template-block.js
osfameron: @mojavelinux I'm confused on how to `include Asciidoctor::Logging` in JS rather than Ruby?

The second link for JS appears to be for choosing the overall Logger for the whole processing, as opposed to calling the logger from an extension.

osfameron: Yikes, I found this on the internets, but feels like there must be a simpler way?

```
  const Opal = global.Opal
  const loggerManager = Opal.module(null, 'Asciidoctor').LoggerManager
  const logger = loggerManager.getLogger()
```

osfameron: Well, I've pushed a commit with that in meantime, but look forward to pointers to improve!

mojavelinux: I looked into it and the use of the Logging module in extensions is not yet well supported yet in Asciidoctor.js.

Here's the low-level JavaScript that would mix in the module:

```js
self.$singleton_class().$include(global.Opal.Asciidoctor.Logging)
```

However, that won't give you access to `getLogger()` and `createLogMessage()` because this doesn't mix in Asciidoctor.js' porcelain interface.

It's probably best to use the Logging module directly:

```js
const Logging = global.Opal.Asciidoctor.Logging

...

Logging.getLogger().error('message')

// or

Logging.getLogger().error(
  Logging.createLogMessage('message', { source_location: parent.getDocument().getReader().$cursor_at_mark() })
)
```

It's also possible to get the logger from the parent or reader:

```js
parent.getLogger()
reader.getLogger()
```

That will give you the ability to log a basic message. However, in order to use `createLogMessage` to add context, you still need to use the Logging module.

home/modules/contribute/examples/helpers/each-openapi.js
```diff
@@ -0,0 +1,164 @@
...
+    
+    // We try to display the parameter as the best possible value,
+    // e.g.
+    //     example -> default -> fallback for the type
+    //
+    // Then, whether we used a fallback like "string" or an example like "Bob",
+    // we format it depending on its type (e.g. surrounding it in quotes etc.)
+
+    function format_type (node) {
+        if (node.type == 'object') return
```
leeming, line 33: (minor?) My JS is rusty (some Typescript tho), but doesn't JS use triple === for boolean truthy, similar to PHP/Perl where strings are booleans, numbers and everything else

mojavelinux: Both work, but you are strongly advised to use `===`. The only time you should use `==` is when comparing against `null` (since that catches `undefined` too).

osfameron:
The each-openapi functionality is used in https://github.com/couchbase/docs-sync-gateway/pull/738 if more context is needed

home/modules/contribute/examples/helpers/each-openapi.js
```diff
@@ -0,0 +1,164 @@
...
+    // we format it depending on its type (e.g. surrounding it in quotes etc.)
+
+    function format_type (node) {
+        if (node.type === 'object') return
+
+        const type = 
+            node.example ??
+            node.default ??
+            types[node.type]?.fallback ??
+            node.type ??
```
sarahlwelton, line 39: Just curious why we're not trying to return the type of `node` first?

osfameron: The logic is:

* an example (e.g. `"An interesting/illuminating value that could go here"`)
* the default (e.g. `1`)
* the fallback (e.g. `true` for boolean values)
* the plain type (e.g. `string`)
* (unknown)

As most nodes have a type (in theory *all of them* but my JSON Schema parsing is a bit basic for `additionalProperties` handling) then that's in some sense the least interesting thing that we can say about them.

PR #677, opened 2023-05-10
tutorials, au revoir - pt I

RichardSmedley:


osfameron (approved):
👍 pending minor query on whether (unrelated) changes to Python and Node version numbers was intended to be committed alongside.

PR #682, opened 2023-05-17
1. C++  2. new releases

RichardSmedley:


etc/nginx/snippets/rewrites.conf
```diff
@@ -1,15 +1,16 @@
 # NOTE rewrite uses 302 redirect by default (assuming replacement does not start with a protocol)
 # current versions
-set $current_version_sdk_c_api '3.3.5';
-set $current_version_sdk_dotnet_api '3.4.5';
+set $current_version_sdk_c_api '3.3.7';
+set $current_version_sdk_cxx_api '1.0.0-dp.5';
```
leeming, line 4: do you want non x.y.z versions being published?

osfameron (approved):
👍 
(not keen on the git commit subject formatting if I'm gonna niggle, a list would be better as a markdown list in the body of the commit message)

etc/nginx/snippets/rewrites.conf
osfameron: should be OK, as this is only declaring a variable, not adding the repo to the playbook

RichardSmedley: As it stands, we're not putting docs up before GA, so the situation won't arise (this is a placeholder),
but it's not impossible that beta docs would be needed here or elsewhere.

PR #685, opened 2023-05-23
Fixing docs-search 

leeming:
1. use pinned working version of glob
2. edit `versioned` and `versionless` helpers to generate working config for Capella (this was broken by the Hosted/Provisioned config)

### DIFF
--- a/docsearch/hbs-helpers/versioned.js
+++ b/docsearch/hbs-helpers/versioned.js
@@ -1,3 +1,11 @@
-const versioned = (components) => components.filter(({ latest }) => (latest || 'master') != 'master')
+// A component with an empty version (defined as `~` in the playbook.yml) 
+// or named 'master' is considered to be "versionless", so the URL generated
+// will be simply e.g. https://docs.couchbase.com/cloud/
+// whereas a "versioned" component might have /current/ or /7.1/ appended.
+
+const versioned = (components) => components.filter(
+    ({ latest }) =>
+        latest != 'master'
+        && latest != '')
 
 module.exports.register = (hbs) => hbs.registerHelper('versioned', versioned)
--- a/docsearch/hbs-helpers/versioned.js
+++ b/docsearch/hbs-helpers/versioned.js
@@ -1,3 +1,3 @@
-const versioned = (components) => components.filter(({ latest }) => latest !== 'master')
+const versioned = (components) => components.filter(({ latest }) => (latest || 'master') != 'master')
 
 module.exports.register = (hbs) => hbs.registerHelper('versioned', versioned)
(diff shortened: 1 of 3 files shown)


osfameron:
Please review: this config has generated Config which has been run successfully against Live (with Jenkins job https://docs-build.sc.couchbase.com/job/Antora/job/docs-search-indexer/job/fix%252Fdocsearch-master/ ) 

Merging this review will add it to the pipeline for *regular* updates to indexing.

Known issues: *Staging* will continue to not be indexed, due to the password protection.

simon-dew (approved):
approved to unblock

osfameron:
fyi @mojavelinux

docsearch/hbs-helpers/versioned.js
```diff
@@ -1,3 +1,3 @@
-const versioned = (components) => components.filter(({ latest }) => latest !== 'master')
+const versioned = (components) => components.filter(({ latest }) => (latest || 'master') != 'master')
```
mojavelinux, line 1: Rather than doing a catch-all for latest here, it would be better to check for the component with the exception by name. That way, nothing surprising happens when a different component shows up.

docsearch/hbs-helpers/versionless.js
```diff
@@ -1,3 +1,3 @@
-const versionless = (components) => components.filter(({ name, latest }) => name !== 'home' && name !== 'tutorials' && latest === 'master')
+const versionless = (components) => components.filter(({ name, latest }) => name !== 'home' && name !== 'tutorials' && (latest || 'master') === 'master')
```
mojavelinux, line 1: Same as above. Address the company by name that requires the exception. Keep the `latest === 'master'` as the else condition.

mojavelinux:
I had a comment about the change to the filter. Other than that, :+1:.

docsearch/hbs-helpers/versioned.js
osfameron: Ah? I thought you've mentioned that `master` and `~` were semantically equivalent, with the latter simply being the preferred modern option?

If another component ends up having a `~` version, wouldn't that also need to be treated as versionless? (e.g. lose the `/current/` path etc.)

mojavelinux: They are semantically equivalent, but only because Antora adds special treatment to the value of "master". The value "master" is still the version value if set that way.

Having a closer look at the code, I see that what it is doing is coercing an empty/undefined value to "master" for the purpose of comparison. I suppose this is reasonable after all.

osfameron: Discussing with @leeming it's definitely a bit unclear!
Will rewrite to something like `((latest === 'master') or (latest === ''))`

PR #690, opened 2023-05-30
Info Banner - Capella Spring Release

osfameron:
This is currently on docs-staging.

Note that Cloud and Cloud Native don't show it, because their landing pages are defined in separate projects.
Will generate a PR for them once approved.

### DIFF
--- a/home/modules/ROOT/partials/info-banner.adoc
+++ b/home/modules/ROOT/partials/info-banner.adoc
@@ -1,0 +1,48 @@
+// This banner partial can be edited to link
+// to marketing or new Docs content as appropriate.
+//
+// This is linked from each of the Landing pages.
+// To *disable* it, simply comment out the text BELOW.
+//
+// See
+// https://couchbasecloud.atlassian.net/browse/AV-56547
+// for the original requirement.
+
+// NOTE: the following CSS will be moved to docs-ui once 
+// iterated and agreed.
+++++
+<style>
+.info-banner {
+    display: block;
+    border: 1px solid #e5e5e5;
+    border-radius: 3px;
+    -webkit-box-shadow: 0 3px 10px rgba(0,0,0,.06);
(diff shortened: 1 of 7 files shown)


osfameron:
@sarahlwelton feel free to push any CSS changes!

(I'll also need to extract them to docs-ui in due course, but inlined here is slightly easier to iterate for POC)

sarahlwelton:
> @sarahlwelton feel free to push any CSS changes!
> 
> (I'll also need to extract them to docs-ui in due course, but inlined here is slightly easier to iterate for POC)

As long as we make sure to undo this - inline CSS is already making me nervous! 😱

sarahlwelton (approved):
I'm picky and want to see the text minorly tweaked for grammar, and I've added my CSS. We can discuss further from here - I'm assuming we didn't hear anything from marketing regarding gradient vs. my style?

osfameron:
Thanks @sarahlwelton - I've not heard back from Marketing, but @bfavini may have info?

I feel like the styling looks more classy (yay) but also feels a bit less noticeable?
Anyway, I'll merge so Beth/marketing can chime in, BUT:

The hover-over button effect is great! I shoulda done that, was lazy of me. **BUT** it wiggles the box by a pixel which is messy. Can you get that fixed, and then we can get it built into staging.

RichardSmedley (approved):
:+1:

home/modules/ROOT/partials/info-banner.adoc
```diff
@@ -0,0 +1,57 @@
...
+    font-size: 14px;
+    font-weight: bold;
+    display: inline-block;
+    border: solid 1px #0074e0;
+    padding: 0.5em 1em;
+    margin-left: 1em;
+}
+.info-banner a:hover {
+	text-decoration: none;
+	border: none;
```
dgnicholson, line 42:
```suggestion
	border: none;
```
I think this might be causing the banner to shift on hover.

PR #691, opened 2023-06-01
Minimal reproduction of Antora logging error

osfameron:
FYI @mojavelinux @leeming, the logging error I mentioned, a minimal reproduction.

It looks like if too many errors are sent to the Antora logger, it doesn't finish logging all of them.

Not sure if this is because messages aren't drained before Antora shuts down, or that the logger is otherwise unresponsive.

Case 1: --log-level=fatal
All the messages are seen:

    ❯ antora minimal.yml --extension=lib/minimal.js --log-level=fatal
    [08:34:26.425] FATAL (andora): Context Started (antora logger)
    Context Started (console.log)
    ...
    Site generation complete!
    Open file:///Users/hakimcassimally/couchbase/docs-site/public in a browser to view your site.
    [08:35:10.242] FATAL (andora): Site Published (antora logger)
    Site Published (console.log)

Case 2: --log-level=info
The on('sitePublished') hook is still run (as we see the console.log) however we DO NOT see the FATAL message via the Antora logger.

    ❯ antora minimal.yml --extension=lib/minimal.js --log-level=info
    [08:36:21.272] FATAL (andora): Context Started (antora logger)
    Context Started (console.log)
    ... (a million errors)
    Site generation complete!
    Open file:///Users/hakimcassimally/couchbase/docs-site/public in a browser to view your site.
    Site Published (console.log)

### DIFF
--- a/lib/minimal.js
+++ b/lib/minimal.js
@@ -1,0 +1,20 @@
+'use strict'
+
+module.exports.register = function ({ playbook, config }) {
+    const logger = this.getLogger('andora')
+    
+    this.on('contextStarted', ({ playbook }) => {
+        logger.fatal('Context Started (antora logger)')
+        console.log('Context Started (console.log)')
+    })
+    
+    this.on('sitePublished', ({ playbook, siteAsciiDocConfig, siteCatalog, uiCatalog, contentCatalog, publications }) => {
+        logger.fatal('Site Published (antora logger)')
+        console.log('Site Published (console.log)')
+    })
+    
+}
+
+
+
+
--- a/minimal.yml
+++ b/minimal.yml
@@ -1,0 +1,50 @@
+antora:
+  extensions:
+  - '@antora/site-generator-ms'
+site:
+  title: Couchbase Capella Staging Docs
+  url: https://docs.stage.nonprod-project-avengers.com
+  start_page: cloud:ROOT:index.adoc
+  keys:
+    google_analytics: GTM-MVPNN2
+git:
+  ensure_git_suffix: false
+content:
+  branches: master
+  sources:
+  - url: .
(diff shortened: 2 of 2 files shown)


mojavelinux:
I believe the issue you're encountering is due to the use of the error level fatal. That should level should only be used when the application has encountered an unrecoverable error and is terminating. The logger (pino) does not guarantee that any log messages will be properly flushed after that.

Regardless, this is not a problem I'm able to reproduce. It's possible this is a result of using an older version of pino due to the fact that Antora has not been upgraded in this build. It's really important that you are at least using Antora 3.1.x instead of 3.0.x because it ensures you are getting the upstream bug fixes.

lib/minimal.js
```diff
@@ -0,0 +1,20 @@
+'use strict'
+
+module.exports.register = function ({ playbook, config }) {
+    const logger = this.getLogger('andora')
```
mojavelinux, line 4: There's no need to pass a value here. "antora" (if spelled correctly) is already the default category.

mojavelinux:
I did a bit more digging here and it appears that you might be encountering [#971](https://gitlab.com/antora/antora/-/issues/971) which was fixed by https://gitlab.com/antora/antora/-/commit/0dd5a8635e9cc487e4130c55f274034276d335e6. That could definitely impact the behavior of the logger.

The package-lock.json file has Antora pinned at 3.0.1, which predated this fix. So if you regenerate the package-lock.json file, it should grab at least Antora 3.0.3, which has this fix. (Though, it still would be better if you upgraded to Antora 3.1.x).

osfameron:
Confirmed this behaviour doesn't manifest after upgrading to Antora 3.1

PR #695, opened 2023-06-06
Upgrading master to use Antora v3.1.2 from v3.0.1

leeming:


leeming:
Incomplete build, but doesn't fail on upgrading

https://docs-build.sc.couchbase.com/job/Andrew%20test/job/Antora%20Upgrade/11/console

PR #696, opened 2023-06-06
AV-55075 Update playbook and Jenkins for @asciidoctor/tabs (Staging)

osfameron:


Dockerfile.jenkins
```diff
@@ -9,6 +9,7 @@ RUN addgroup -g $GROUP_ID jenkins && \
     yarn global add --ignore-optional --silent git+https://gitlab.com/opendevise/oss/antora-site-generator-ms#as-extension && \
     yarn global add --ignore-optional --silent asciidoctor-kroki && \
     yarn global add --ignore-optional --silent asciidoctor-external-callout@1.2.0 && \
+    yarn global add --ignore-optional --silent https://yarn.pm/@asciidoctor/tabs && \
```
mojavelinux, line 12: The Asciidoctor Tabs dependency should not be prefixed with https://yarn.pm/.

osfameron:
Indeed! I don't know yarn at all, was trying to copy/paste from the yarn modules site to get it right, and misunderstood what to paste... (it's fixed in a subsequent commit after this PR)

Do you remember the reason we use npm locally and yarn on the builds?
Is there any disadvantage to standardising on one OR the other for both?

mojavelinux:
Antora's container uses yarn, so it's currently required since this container extends that one. In Antora, we're in the very slow process of moving away from Yarn, but we've not had ample time to complete the transition. Hence, we're stuck using yarn in one case and npm in another. Though, in the end, it really doesn't matter what is used since they do exactly the same thing.

PR #697, opened 2023-06-07
Antora3.1 archive

leeming:


mojavelinux:
FYI, Antora 3.1.4 release is now available with a fix for the credentials issue when git@ is used in the content source URL. You may decide you don't want that anyway, but just letting you know.

See https://docs.antora.org/antora/latest/whats-new/#antora-3-1-4

PR #699, opened 2023-06-13
AV-57335 POC for analytics include using collector

osfameron:
To simplify process for Analytics team, we can use https://gitlab.com/antora/antora-collector-extension/

NOTE: this extension is currently Alpha, will have to discuss with OpenDevise.

This PR will add the Couchbase fork of AsterixDB and the Collector extension.
A subsequent PR on the AsterixDB repo will provide a command to aggregate sources.
https://github.com/couchbase/asterixdb/pull/2

The .adoc file in this PR includes one of these files, using the Markdown block (modified to cope with the use of HTML in AsterixDB sources)

### DIFF
--- a/antora-playbook.yml
+++ b/antora-playbook.yml
@@ -5,6 +5,7 @@ antora:
   extensions:
   - '@antora/site-generator-ms'
   - ./lib/embargo.js
+  - '@antora/collector-extension'
 
 site:
   title: Couchbase Docs
--- a/antora-playbook.yml
+++ b/antora-playbook.yml
@@ -73,9 +73,14 @@ content:
     start_path: docs
   # - url: https://git@github.com/couchbase/docs-connectors-talend
   - url: https://github.com/couchbase/docs-connectors-talend
+
   # - url: https://git@github.com/couchbase/docs-analytics
   - url: https://github.com/couchbase/docs-analytics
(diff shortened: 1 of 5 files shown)


osfameron:
@mojavelinux I've wanted to test this extension out for a while since @RayOffiah mentioned it, but was blocked on the Antora 3.1 upgrade.
This POC was *much* easier than expected, this is a really powerful feature for exactly this kind of issue (3rd party repos where we have minimal control over the content structure.)

I've raised one minor issue on GitLab -- which may just be a bad expectation on my part.

The extension is currently marked Alpha. Do you recommend that we proceed beyond this initial demo to a fully fleshed out POC?

mojavelinux:
> The extension is currently marked Alpha. Do you recommend that we proceed beyond this initial demo to a fully fleshed out POC?

I would be cautious. Alpha means that the extension could change at any time and that we might even complete redesign it. There are no guarantees. However, if you are prepared for a potential migration in the future, then you can use the version that is released and either stay with that or do any necessary migration when 1.0.0 is available.

osfameron:
OK. Assuming that we're using the package *as intended*, then it feels like even a substantial redesign should be manageable with a reasonable effort (update configuration, update paths, change how the command payload is called etc.)

I'm ok with that from tooling side, @simon-dew what do you think as stakeholder?

PR #700, opened 2023-06-13
AV-54052: fix package-lock

osfameron:
Looks like we need to have site-generator at 3.1, for `npx antora` to Just Work.
Unsure if we should now also remove the site-generator-ms entry? Ping @mojavelinux

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,3 @@
+fund=false
+lockfile-version=3
+omit=optional
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,13 +1,13 @@
 {
   "name": "docs-site",
-  "lockfileVersion": 2,
+  "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "docs-site",
       "dependencies": {
         "@antora/site-generator-ms": "git+https://gitlab.com/opendevise/oss/antora-site-generator-ms#as-extension",
-        "@asciidoctor/tabs": "^1.0.0-beta.5",
+        "@asciidoctor/tabs": "1.0.0-beta.5",
         "antora": "~3.1",
         "asciidoctor-external-callout": "~1.2.0",
         "asciidoctor-kroki": "0.15.4",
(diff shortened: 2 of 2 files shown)


mojavelinux:
Adding the site generator package shouldn't be necessary as it is already pulled in by the antora package.

Yes, you still need site-generator-ms because it's not a site generator anymore, but rather a collection of Antora extensions which are still being used. The goal is to eventually be able to use only packages from the Antora project (such as Antora Atlas and Antora Navigator), but those packages are not developed enough to replace the extensions in site-generator-ms.

Since this build is using a package.json file, using `npx` still requires running `npm ci` first. When `npx` is run without a node_modules directory, it tries to only use the packages that are specified in the command, which are not enough to build this site.

osfameron:
Ah! This is because `npx ci` gives a cleaner install than `npm i` ?

Having done the latter, my colleagues were finding that `npx antora -v` gave the expected 3.1.x, while `npx antora some-playbook` gave the `FATAL (antora): Generator not found or failed to load. Try installing the '@antora/site-generator' package.` message.

mojavelinux:
> This is because npx ci gives a cleaner install than npm i ?

`ci` installs dependencies listed in package-lock.json so that every installs exactly the same versions. Only the person managing this file needs to run `i`, which is what generates this file.

See:

* https://docs.npmjs.com/cli/v9/commands/npm-ci
* https://docs.npmjs.com/cli/v9/commands/npm-install

osfameron:
Ah! Thanks, that's been a mystery... I need to learn Node.js properly...
Looks like we accidentally downgraded the lockfile last year. I'll keep an eye on this from now on 👍

osfameron:
Nice. Thanks, all done, will merge after review from colleagues, and read up on Node/npm...

.npmrc
```diff
@@ -0,0 +1,3 @@
+fund=false
+lockfile-version=3
```
osfameron, line 2: This is the key line: we pin to version 3

osfameron: Reviewing package-lock.json is tricky (Github can't cope with it... command-line git is fine)
You could note that:

line 3: `"lockfileVersion": 3,`
line 186:
```
    "node_modules/@antora/site-generator": {
      "version": "3.1.4",
```
so even though we've not, in the end, added site-generator explicitly, it is pulled in by the version 3 lockfile.

osfameron: For bonus points, check out this PR, and then

```
npm ci
npx antora -v
```
and check that both cli and site-generator are 3.1.4

PR #704, opened 2023-06-16
AV-56547: Revert Banner

osfameron:
revert the Capella Spring Banner.

NB: 
* the calling points are left in place, so it should be trivial to edit this file to re-add the banner at any point in future

* we remove the inline CSS, which will now be added to docs-ui in PR https://github.com/couchbase/docs-ui/pull/167


dgnicholson (approved):
LGTM

PR #706, opened 2023-06-21
SDK Workaround

RichardSmedley:
Time-saving fix to legacy 6.5 includes in SDK 3.1.

In later versions a variable ensures the correct include::, but not in 3.1.


osfameron (approved):
👍

PR #710, opened 2023-07-04
Adding initial example of github actions workflow that is required to…

leeming:
… be merged before testing, grumble


osfameron:
OK, feel free to create a new repo in couchbaselabs (I think you'll have permissions, shout at me if not)

PR #712, opened 2023-07-10
Fix mobile tutorials redirects

osfameron:
Via https://www.couchbase.com/forums/t/documentation-of-travel-sample-app-not-found/36258/2

We redirect /tutorials/ to developer.couchbase.com but the mobile tutorials were never adopted.

Until these are also adopted, this commit does `break` for known tutorials, to prevent the following redirect.

Test that e.g. https://docs-staging.couchbase.com/tutorials/todo-app/ is 404 (rather than 302 redirect).

The tutorials are published to live, e.g. see
https://github.com/couchbase/docs-site/blob/master/antora-playbook.yml#L130-L157 but not staging, hence the 404.


franrob-projects (approved):
LGTM

RichardSmedley (approved):
:+1:

PR #725, opened 2023-08-21
Patching jenkinsfile to run global sitemap script

leeming:


osfameron (approved):
Looks good to me.
(If in doubt, then cherry-pick to staging first, or we can run and just revert it if any issues)

PR #739, opened 2024-01-30
[WIP] DOC-11819: remove playbook references to tutorials.

ElliotFrancisHunter:
This is an attempt to fix the following ticket: https://issues.couchbase.com/browse/DOC-11819

https://docs.couchbase.com/tutorials/userprofile-query-android/userprofile_query.html

https://docs.couchbase.com/tutorials/mobile-travel-tutorial/swift/installation/travel-mobile-app.html

On triaging: https://issues.couchbase.com/browse/DOC-10798

I discovered a few orphan pages, it is my understanding that tutorials have moved to the developer advocacy side which would explain the broken appearance of the pages (screenshots attached). No nav bars being one reason.

### DIFF
--- a/antora-playbook.yml
+++ b/antora-playbook.yml
@@ -22,7 +22,6 @@ site:
         { "title": "Autonomous Operator", "components": ["operator"] },
         { "title": "CMOS", "components": ["cmos"] },
         { "title": "SDKs", "startPage": "home::sdk.adoc", "components": ["*-sdk", "cxx-txns", "elasticsearch-connector", "kafka-connector", "spark-connector", "tableau-connector", "power-bi-connector", "sdk-extensions"] },
-        { "title": "Tutorials", "startPage": "tutorials::index.adoc", "components": ["tutorials"] }
       ]
 git:
   ensure_git_suffix: false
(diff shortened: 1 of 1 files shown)


antora-playbook.yml
```diff
@@ -124,18 +123,8 @@ content:
...
-  - url: https://github.com/couchbaselabs/mobile-travel-sample
-    branches: [master]
-    start_path: content
   # - url: https://github.com/couchbaselabs/hotel-finder-react-native
   # - url: https://github.com/couchbaselabs/hotel-lister-cordova
-#  - url: https://github.com/couchbaselabs/tutorials-contrib
-  - url: https://github.com/couchbaselabs/mobile-training-todo
-    branches: tutorials
-    start_path: content
   - url: https://github.com/couchbaselabs/UniversityLister-Android
```
osfameron, line 128: While you're here, I think the following modules are all tutorial content too.

And looking at the names vaguely reminds me of a chat a while back with @RichardSmedley... didn't you find some reason for which this content was spared, as it's still required by PM / not yet adopted by Dev Advocacy?

antora-playbook.yml
```diff
@@ -22,7 +22,6 @@ site:
         { "title": "Autonomous Operator", "components": ["operator"] },
         { "title": "CMOS", "components": ["cmos"] },
         { "title": "SDKs", "startPage": "home::sdk.adoc", "components": ["*-sdk", "cxx-txns", "elasticsearch-connector", "kafka-connector", "spark-connector", "tableau-connector", "power-bi-connector", "sdk-extensions"] },
```
osfameron, line 24: I think you need to remove the trailing `,` here.

(Though this is in yaml, it's actually a *string* representation of JSON for, er... historical reasons, I imagine...)
And I think JSON spec strictly speaking doesn't allow trailing commas, even though JS itself does, 🤷 hey ho.

osfameron (commented):
I think this looks good with 1 technical tweak, but let's see what other reviewers bring to table!

antora-playbook.yml
RichardSmedley: You may be right, but I can't remember - DA are at SKO this week, so we may not get a quick answer on that.

PR #740, opened 2024-02-08
migrate build for archive site from Antora 2 to Antora 3

mojavelinux:
- document that Node.js 16 is now the minimum required version of Node.js
- switch to antora package and upgrade it to ~3.1
- switch to extension-based @antora/site-generator-ms (as-extension branch)
- register @antora/site-generator-ms as Antora extension
- remove references to outdated xref validator
- add lib/exclude-private-cotnent-sources-extension.js for testing site without access to private repos
- refactor Asciidoctor extensions to work with Asciidoctor 2
- refresh dependency lock file and use version 3
- @asciidoctor/tabs is not compatible with the archive stylesheet, so lib/tabs-block.js is still required


mojavelinux:
Note that fetch_concurrency is now set to 1 by default, so there's no need to set it in the Antora playbook.

PR #743, opened 2024-02-26
Style guide update

osfameron:
https://preview.docs-test.couchbase.com/styleguide/home/contribute/index.html as discussed in team meeting


sarahlwelton (commented):
I guess the couple questions I have right now... 

1. How will we build to update this site going forward?
2. Are we going to stay on the preview build side of docs-site, or is this just showing what the new contribute guide will look like on live? 
3. If it's the second half of point 2, does that mean that it will automatically build nightly? 

Just want to know how I can get changes into the style guide going forward :)

osfameron:
Yes, to merge into live!
I should have pushed a commit changing the main playbook?’(Will check and push if not tomorrow)

sarahlwelton (approved):
Honestly, unless anyone else has objections, I think I'm satisfied with this.

PR #746, opened 2024-03-11
Updated .NET SDK version to 3.5.0

julia-cb:
Updated .NET SDK version to 3.5.0 in order to publish updated release notes.


simon-dew (approved):
Approving to unblock. LGTM, but with the caveat that I know nothing about the .NET SDK update.

PR #747, opened 2024-03-11
Updated file with .NET SDK release/3.5 branch

julia-cb:
Added .NET SDK branch release/3.5 to file.


simon-dew (approved):
Approving to unblock. LGTM, but with the caveat that I know nothing about the .NET SDK update.

PR #748, opened 2024-03-18
DOC-11976: server 7.6 GA

osfameron:
Add 7.6 branches (or `trinity` for those repos that use that convention).

Update UI bundle for Metrics display.


RichardSmedley (approved):
LGTM

(NOTE, I will be updating this file for Ruby SDK shortly - so you may need to rebase then.)

simon-dew (approved):
👍 LGTM

PR #753, opened 2024-04-09
AV-76422: Add attribute token for columnar-flavored SQL++

lamagnifica:
See https://github.com/couchbaselabs/docs-columnar/pull/51.


RayOffiah (approved):
Yup, that should work!

PR #755, opened 2024-04-22
DOC-12133 update page-embargo functionality

osfameron:
Via Simon Dew, a page with page-aliases breaks page-embargo.

Looking at original commit, we were manually deleting page.pub etc. because the contentCatalog method was only present in Antora 3.1.1 and onwards. see 13190cefc6891db7500d086a1ddd09f13bb64bab

As we're now upgraded, we can substitute this in.


simon-dew (approved):
LGTM 👍 AFAICT

I like the way that the solution was right there already, waiting to go!

simon-dew:
UPDATE: Tested locally and this fix works.

PR #756, opened 2024-04-30
Update Java and Scala SDK versions to 3.6.2 and 1.6.2

julia-cb:
Updated Java and Scala versions to match new versions coming out.


RichardSmedley (approved):
:+1:

PR #759, opened 2024-05-08
SEO-changes-map-302-to-301

RayOffiah:
Test change to map 302 redirections to 301, which should allow them to be indexed by search engines. Testing on Staging first, in case it breaks.


RayOffiah:
Following conversation with @Hakim, redid the change to just target the tutorials section.

osfameron (approved):
Looks good! 👍

PR #763, opened 2024-05-17
The Big Writer's Guide Update

sarahlwelton:
See a preview here: https://preview.docs-test.couchbase.com/home/contribute/index.html 

I spent this week going through and trying to clean up some of the key parts of the writer's guide, to better support some of the new joiners that we'll be having go through getting set up with Antora. 

Feel free to call out other changes you'd like to see.


osfameron (approved):
This looks fabulous and very thorough.
I've skimmed through, and can't see any reason not to 👍 it -- I've not looked at it closely, and the preview link seems to be wrong?
But it's undoubtedly better than what we had, and I think we should merge and iterate on it.

(Happy to chat if there's any specific sections you'd like critique on though!)

One thought: I've found taking people through github Personal Access Tokens confusing... ssh keys work great for me, and `gh repo clone` is, honestly, much much easier than anything else (though it involves now having *2* command line commands, so perhaps your mileage may vary).
That's just to suggest you try that out at some point, and see if you think it's worth documenting in parallel/replacing?  Happy to contribute that if helpful.

PR #767, opened 2024-05-27
[AV-60739] Attributes for Day 0 Docs

sarahlwelton:
Adding necessary {db} and {cptl-db} attributes to playbooks for building docs after Day 0 updates. 

Will make the future transition onto "operational database" much easier, if that ends up being the final decision.

### DIFF
--- a/antora-playbook-staging.yml
+++ b/antora-playbook-staging.yml
@@ -183,6 +183,8 @@ asciidoc:
     sqlpp: SQL++
     sqlpp_url: https://www.couchbase.com/products/n1ql
     cbpp: Couchbase++
+    db: cluster
+    cptl-db: Cluster
     kroki-server-url: http://3.91.133.254:9500
     kroki-fetch-diagram: true
   extensions:
--- a/antora-playbook.yml
+++ b/antora-playbook.yml
@@ -189,6 +189,8 @@ asciidoc:
     sqlppc: SQL++ for Capella columnar
     sqlpp_url: https://www.couchbase.com/products/n1ql
     cbpp: Couchbase++
+    db: cluster
+    cptl-db: Cluster
     kroki-server-url: http://3.91.133.254:9500
     kroki-fetch-diagram: true
     # the url-issues and url-issues-* attributes configure the URLs for the inline jira macro


antora-playbook-staging.yml
```diff
@@ -183,6 +183,8 @@ asciidoc:
     sqlpp: SQL++
     sqlpp_url: https://www.couchbase.com/products/n1ql
     cbpp: Couchbase++
+    db: cluster
+    cptl-db: Cluster
```
osfameron, line 187: As someone who has enjoyed figuring out what attributes like `{sgw-te}` and `{cblFrmWk}` in Mobile mean, I'm excited about everyone else getting the opportunity to participate in the fun!

(Don't mean to snark, I've perpetrated enough of these myself... anyway, seeing it in a PR makes me wonder if we should try to do something about it and maybe come up with some best practices ongoing.)

e.g. What does `cptl` mean? Can we spell it out in full?
Or is it worth adding YAML comments, like e.g.:

`db: cluster   # the name of a "database-like-thing"`

RichardSmedley: Could be listed in the README?

(I'm in the process of doing just that for SDK Docs' own growing list of attributes)

sarahlwelton: I was trying to keep it short for the sake of easy typing in running doc. 

If we'd rather spell it out for explicitness, I'm okay with that.

lamagnifica: an inline comment would be wonderful from my PoV

lamagnifica:
@sarahlwelton do we need to start with tokens for both database and cluster? or product-specific tokens?
Capella uses cluster in the APIs and database in the UI to mean the same object. 
Capella columnar services uses cluster for, essentially, the same object again, in both the API and the UI.

Rumor is that eventually, in Capella database will be replaced with operational database, but the API will continue to use cluster. 
In sum, "cluster" is a valid term, and "database" is a valid term, and the only potential change coming is "database" to "operational database". 
I think that adding only a single token now for cluster, and naming it {db},  is going to be problematic.

PR #768, opened 2024-06-12
AV-79965 point at docs-tableau after extraction

osfameron:


simon-dew (approved):
LGTM 👍 

Are you going to create a ticket to remove the striping from the versions, or do you want me to do that?

simon-dew:
> Ah, I'm not sure I understand what you mean there, may be best for you to do it and explain what's left to do? (Happy to then get involved in fixing it if I can help)

NP, I'll create the tickets

PR #771, opened 2024-07-01
AV-80236 new connectors repos

osfameron:


RayOffiah (approved):
Yup, looks good to me! 👍🏾

RichardSmedley (approved):
LGTM :+1: 
(we can rename the main branches on the new repos, once we GA new branches in the future - and socialise that with engg.)

PR #773, opened 2024-07-05
quickfix: make fetch default in playbook

osfameron:
Make Antora fetch remote updates to repositories:
  https://docs.antora.org/antora/latest/playbook/runtime-fetch/

This will be a No-Op in most builds as:

	* for Github Actions/Jenkins build, the environment is generated completely fresh every time, and Antora will always fetch the latest content in this case.

	* most writers working locally will point to local repos, where this statement does nothing

So... what's the point? Well, the master playbooks are a resource that writers often copy, and having this as a default will help in the case that:

	* Locally, you build with a remote repo, and don't update changes, leading to confusion and the gnashing of teeth.

(Those writers who use that workflow regularly have already added this to their playbooks, or have aliased antora to always use `--fetch`. But it confuses the rest of us who don't do this regularly.)

### DIFF
--- a/antora-playbook-staging.yml
+++ b/antora-playbook-staging.yml
@@ -198,3 +198,5 @@ ui:
     url: https://github.com/couchbase/docs-ui/releases/download/preserve-hash-on-version-change-1/ui-bundle.zip
 output:
   dir: ./public
+runtime:
+  fetch: true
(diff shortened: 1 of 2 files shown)


RayOffiah (approved):
LGTM 👍🏾

simon-dew (approved):
LGTM 👍 &mdash; I always run antora with `--fetch` anyway. 

Not tested, but the playbook format looks correct, cf. [Fetch Updates](https://docs.antora.org/antora/latest/playbook/runtime-fetch/)

PR #778, opened 2024-07-26
Update sd-capella-preview

simon-dew:


etc/nginx/snippets/rewrites.conf
```diff
@@ -1,17 +1,17 @@
...
-set $current_version_sdk_java_api '3.5.0';
-set $current_version_sdk_jvm_core_api '2.5.0';
-set $current_version_kotlin_client_api '1.2.0';
-set $current_version_sdk_nodejs_api '4.2.8';
-set $current_version_sdk_php_api '4.1.6';
-set $current_version_sdk_python_api '4.1.9';
-set $current_version_sdk_ruby_api '3.4.5';
-set $current_version_sdk_scala_api '1.5.0';
+set $current_version_sdk_c_api '3.3.12';
+set $current_version_sdk_cxx_api '1.0.0-dp.15';
```
RichardSmedley, line 4: If these are being updated, is there any reason not to update them to latest to match prod?

antora-playbook.yml
```diff
@@ -77,48 +91,48 @@ content:
...
-    branches: [release/7.2, release/7.1.2, release/7.1, release/7.0, release/6.6, release/6.5]
+    branches: [temp/7.5, release/7.2, release/7.1.2, release/7.1, release/7.0, release/6.6, release/6.5]
   - url: https://github.com/couchbase/docs-sdk-c
-    branches: [release/3.3, release/3.2]
+    branches: [release/3.3]
   - url: https://github.com/couchbase/docs-txn-cxx
     branches: [release/1.0]
   - url: https://github.com/couchbase/docs-sdk-dotnet
-    branches: [release/3.4, release/3.3, release/3.2]
+    branches: [release/3.5, release/3.4, release/3.3]
```
RichardSmedley, line 112: These are all one dotminor behind, and match 7.6, rather than 7.6.2 - is that deliberate, or should they be updated?

etc/nginx/snippets/rewrites.conf
simon-dew: Ignore this PR, it was caused by a mixup between forks of docs-site

antora-playbook.yml
simon-dew: Take off and nuke the site from orbit

PR #786, opened 2024-08-19
Update antora-playbook-staging.yml for operator 2.7 / 2.8

osfameron:
This must not be merged before confirmation from @malarky that master branch has been moved to track 2.8!

e.g. see https://github.com/couchbase/couchbase-operator/blob/master/docs/user/antora.yml#L3


malarky (approved):
This is good to go now, thank you!

PR #798, opened 2024-11-18
Cloud JIRA link update

RichardSmedley:
DOC-12723


RichardSmedley:
This attribute seems like over-engineering - 
but just fixing for move to cloud JIRA for now.

osfameron (approved):
Oh, is there no longer a redirect from the old link? 😱

PR #799, opened 2024-11-18
DOC-12742 quieten jira: macro

osfameron:


lib/inline-jira-macro.js
```diff
@@ -43,7 +43,6 @@ function register (registry, context) {
     return accum
   }, { __DEFAULT__: 'https://issues.couchbase.com/browse' })
```
simon-dew, line 44: Bit puzzled by this one. If we're removing this extension from the playbook, then why bother updating the extension so it doesn't output to the log? But on the other hand, if you _are_ updating this extension, shouldn't you at least update the default URL?

```suggestion
  }, { __DEFAULT__: 'https://jira.issues.couchbase.com/browse' })
```

simon-dew (commented):
Just some questions:

lib/inline-jira-macro.js
osfameron: Removing the output will mean that anyone with a *local* playbook that still references the extension will now remove the extra noise once they update docs-site.

(If we simply deleted the extension, then it would break local builds.)

As to why not update the default URL... er... no real reason, I just didn't notice, and as nobody uses it, it doesn't seem that important?

simon-dew: Okay, fair enough on the first two points!

simon-dew (approved):

PR #804, opened 2025-01-08
Superset Release 1.0.5

zeba-shireen:
Superset Release 1.0.5


osfameron:
@RichardSmedley hah, yes, we realised this after I suggested to @zeba-shireen to add the live PR...

Anyway, we *have* tested: see https://docs-staging.couchbase.com/home/sdk.html
built via `hakim-test` branch which points to my `preview-build` docs-site branch which folds in Zeba's staging test

(sorry for the confusion - I want staging on that ref for the demo at all-hands later)

PR #807, opened 2025-02-10
Warn on undefined repo in ext.preview.sources

osfameron:
We already warned when the *current* content directory is not defined in antora-playbook.preview.yml.
But we didn't trap this error nicely in the ext.preview.sources case.

This commit extracts the `notfound` function so that both paths can throw a consistent error. (Of course the new case doesn't have all the information, so is a little hand-wavey...)


RayOffiah (approved):
Nice! 👍🏾

PR #810, opened 2025-02-25
improve each-openapi for arrays

osfameron:
* complex arrays are treated as with `handleObject` but with square `[]` rather than curly `{}` braces.
* simple arrays are inlined `["string"...]` in a slightly hacky way... 🙈 
* the handler callbacks are all renamed to `handleX` for clarity


simon-dew (approved):
LGTM 👍

PR #816, opened 2025-04-29
DOC-13225 override preview config with flag

osfameron:
(Only for local previews - could add to remote, but I do like the current case where all previews are easily mapped to the config that created them, does make it easier to manage)

@simon-dew test with

```
# in docs-site
git fetch
git checkout DOC-13225-preview-config

# in your repo
cp preview/HEAD.yml preview/SIMON.yml
# edit config as required

preview -c SIMON
```


simon-dew (approved):
(Unsuccessfully fights temptation to make "what's upfind?" joke)

LGTM 👍

PR #823, opened 2025-05-28
DOC-13194, DOC-12510: improve openapi/JSON Schema handling

osfameron:
* infer type: object if appropriate Though it would be convenient for us if all nodes were tagged, redocly doesn't declare that as an issue with the spec, so we should handle it.
This was preventing `users`, `roles` etc. from being traversed:

image

* descend into `allOf`
We now see the new `index`:
image

NEXT: add tests (long overdue, lack was hampering me in implementing this)

### DIFF
--- a/test/test-build.js
+++ b/test/test-build.js
@@ -13,7 +13,7 @@ describe('antora build with extensions', function () {
       '--log-level', 'fatal', 
       '--to-dir', 'test/build',
       'antora-playbook-test.yml'],
-    {'stdio': 'inherit'})
+      {'stdio': 'inherit'})
   })
 
   const files = 


lib/helpers/lib/each-openapi-module.js
```diff
@@ -0,0 +1,202 @@
...
+    }
+    
+    const via_additionalProperties =
+        node.additionalProperties ?
+            [[
+                `{${node.additionalProperties["x-additionalPropertiesName"] || "additionalProperties"}...}`,
+                format_additionalProperties(node.additionalProperties) ]]
+            : []
+
+    const via_allOf = node.allOf?.flatMap(a => Object.entries(a.properties)) || []
```
simon-dew, line 140: Any plans to do `oneOf`, `anyOf` as well?

simon-dew (commented):
Sorry, I keep leaving reviews as you're working on it.

osfameron:
> Sorry, I keep leaving reviews as you're working on it.

my bad, should've marked it draft. I think it's ready nowish 👍

lib/helpers/lib/each-openapi-module.js
torcolvin: We don't use oneOf in this specification of configuration because we'd really have to do some juggling to make this work for JSON unmarshalling. That said, I could see it ending up being used by accident due to includes.

The places oneOf is used by SG are places where we return different output depending on parameters. e.g. https://docs.couchbase.com/sync-gateway/current/rest_api_admin.html#tag/Database-Management/operation/get__all_dbs

Screenshot 2025-05-29 at 9 57 50 AM

I am unclear on how this is used, is this code used at runtime and run client side javascript, or does this generate static values? If this generates static values, is it possible to just force an error that would bubble up when you build the documentation?

lib/helpers/lib/each-openapi-module.js
```diff
@@ -0,0 +1,202 @@
...
+
+        if (! node.type) {
+            if (
+                node?.additionalProperties?.type === 'object' || 
+                node?.allOf?.[0]?.type === 'object'
+            ) {
+                node.type = 'object'
+            } else {
+                // node.type = 'unknown'
+                console.warn(`No type for node ${path}`)
```
torcolvin, line 167: As above, if this is running on client side javascript for the user, I'm not sure if this is helpful.

If this is run as part of a build process for documentation, is it possible to make this error and bubble back up to the builder of the documentation website? If this is done at runtime, is it possible to run this code at build time to bubble the code back up to the user?

lib/helpers/lib/each-openapi-module.js
```diff
@@ -0,0 +1,202 @@
...
+// to handle the tree as appropriate
+// this can be used to:
+//    * generate a tree (as per process_toc)
+//    * flatten the tree into a list (as per get_items)
+function iterate_properties (node, handleLeaf, handleObject, handleArray) {
+    
+
+    function recurse(node, path) {
+
+        if (! node.type) {
```
torcolvin, line 159: I looked deep into the specification https://swagger.io/specification/ and I may not have done this correctly, but I can't see that https://swagger.io/specification/#schema-object a schema object _ever_ requires a type to be declared. I played with redocly a bit, and it seems like if you don't specify a type, it looks like it picks `object` as the type in the display, but I couldn't find a specification or code for this.

I added some code here https://github.com/couchbase/sync_gateway/pull/7548 to add any missing `type:` and it seems like it is always `type: object`. I don't know this is necessarily true.

It might be interesting to look at swagger code base to see if they implicitly handle type as object unless specificed.

lib/helpers/lib/each-openapi-module.js
```diff
@@ -0,0 +1,202 @@
...
+            if (
+                node?.additionalProperties?.type === 'object' ||
```
torcolvin, line 161: I didn't know this before looking at it, but `additionalProperties` I think is always an object. It can actually be `additionalProperties: true` which is equivalent to `additionalProperties = {}` https://swagger.io/docs/specification/v3_0/data-models/dictionaries/#free-form-objects but that's not used in SG.

I think that this should be `node?.additionalProperties !== undefined`  rather than `node?.additionalProperties?.type === 'object'` to presume this is an object.

lib/helpers/lib/each-openapi-module.js
```diff
@@ -0,0 +1,202 @@
...
+//    * flatten the tree into a list (as per get_items)
+function iterate_properties (node, handleLeaf, handleObject, handleArray) {
+    
+
+    function recurse(node, path) {
+
+        if (! node.type) {
+            if (
+                node?.additionalProperties?.type === 'object' || 
+                node?.allOf?.[0]?.type === 'object'
```
torcolvin, line 162: This has the problem above that it _happens_ that Sync Gateway has declared type as object in the use of `allOf` but that's not guaranteed unless Sync Gateway enforces it. `allOf` is basically used to do inheritance in openAPI, so it would also be possible that `allOf` has multiple entries with different types, or missing types. This is being pedantic, here.

osfameron: If we need it now *or* are worried that it'll bite us in future, then we can do better.
I don't really understand the allOf semantics though.
If you can supply:
* a data-structure to generate (and any comments on what output you expect)
* a failing test-case
then I could look at fixing?

lib/helpers/lib/each-openapi-module.js
```diff
@@ -163,8 +163,9 @@ function iterate_properties (node, handleLeaf, handleObject, handleArray) {
             ) {
                 node.type = 'object'
             } else {
+                node.type = 'object' // default to object?
```
torcolvin, line 166: Simplify this section to be:

```
if (!node.type) {
       node.type = 'object'
}
```

Suggested inline comments:

```
// If type is not specified on a schema object, assume it is a n object. This behavior seems to match redocly behavior.
// It is necessary to explicitly set type on object since format_type function will examine the type attribute on the node object.
```

I have some bias for my name not being mentioned in code, but I would accept it, but I think this whole if block can now be simplified.

lib/helpers/lib/each-openapi-module.js
simon-dew: My understanding is, if you don't specify the type, the type is undefined. OpenAPI Generator lists the schema as "Any type".

PR #825, opened 2025-06-10
DOC-13292: Sync Gateway 3.3

simon-dew:
Add Sync Gateway 3.3 to docs staging build


simon-dew:
> btw have we talked to PM about removing the EOL 2.8 pages?

Good point! I don't think we've mentioned it. I'll try to remember to bring it up

PR #826, opened 2025-06-13
EOL SDKs & evergreen API redirects

RichardSmedley:


osfameron (approved):
This looks fine: one issue we now have is that the build will now silently push the changes even if they could cause nginx to fail in future. 

Therefore please merge, build and *test* that the new rewrites are in place as expected, and shout at me if not
(so we can revert/investigate)

(This is on my TODO to improve...)

PR #828, opened 2025-07-02
Add NGINX redirect rules for erroneous API pages and test configuration

RayOffiah:
Added rules to trap and redirect the pages for API `Enumerations` and `Structures`.

### DIFF
--- a/etc/nginx/snippets/rewrites.conf
+++ b/etc/nginx/snippets/rewrites.conf
@@ -242,3 +242,14 @@ location ^~ /admin/admin/ {
     # final redirect to archives
     rewrite ^/admin/(.*)$ $scheme://docs-archive.couchbase.com/docs-3x/$1 permanent;
 }
+
+# Send some of the erroneous API pages to their correct pages
+
+location ~* ^/mobile/([^/]+)/([^/]+)/Enumerations\.html$ {
+    return 301 /mobile/$1/$2/Enums.html;
+}
+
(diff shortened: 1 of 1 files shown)


etc/nginx/snippets/nginx.conf
```diff
@@ -0,0 +1,27 @@
+# Used with a docker container for localised testing.
```
osfameron, line 1: Presumably the nginx.conf is just for your local testing? I don't think our deploy will do anything with it.

Perhaps rename to `nginx.conf.example` and extent the comment to explain what it's used for.

osfameron (commented):
doesn't seem to work, e.g

* 404 https://infratest.docs-test.couchbase.com/mobile/2.8.4/couchbase-lite-swift/Enumerations.html
* 404 https://infratest.docs-test.couchbase.com/mobile/2.8.4/couchbase-lite-swift/Structures.html

Looking at our rewrites, we don't use `~*` at all... I wonder if we could use the approach used elsewhere, something like:

```
location ^~ /mobile/ {
    rewrite ^/mobile/([^/]+)/(couchbase-lite-swift)/Enumerations.html$ /mobile/$1/$2/Enums.html redirect; 
    rewrite ^/mobile/([^/]+)/(couchbase-lite-swift)/Structures.html$ /mobile/$1/$2/Structs.html redirect; 
}
```

(Not tested 🙈)

RayOffiah:
Works locally, but so did the last attempt.

PR #837, opened 2025-08-14
Removing unnecessary topnav groupings

RichardSmedley:
+ bundle update and new Kafka


RichardSmedley:
> * bundle update and new Kafka

Oops, I meant C++
Kafka was already on staging :facepalm:

osfameron (approved):

PR #840, opened 2025-09-01
DOC-13521 calculate staging playbook

osfameron:
Rationale: as well as the staging playbook, we now have to manage:

* the template preview playbook
* 1+ playbooks for chatbot
* escrow playbook

All of which need to be kept in sync with the main playbook, preserving just important differences.

As the staging playbook is the most high-profile one with likely the most complexity in setting it up, I'm starting with that, to check that we can do this fully enough to support the requirement.

(The preview one is currently supported by the more straight-forward scripts/patch-preview-playbook, but could use the same code.)

### DIFF
--- a/antora-playbook-staging.yml
+++ b/antora-playbook-staging.yml
@@ -1,13 +1,66 @@
-# npx antora --clean --fetch antora-playbook.yml
-# or
-# npx antora antora-playbook.yml
+######
+# WARNING: DO NOT EDIT BY HAND
+# (it will get overwritten)
+#
+# Instead
+#
+#  * edit antora-playbook-staging.diff.yml with just the *differences* from the main playbook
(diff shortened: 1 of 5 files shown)


antora-playbook-staging.yml
```diff
@@ -41,207 +94,266 @@ site:
...
+    - url: https://github.com/couchbase/docs-sdk-nodejs
+      branches:
+        - temp/4.5
+        - temp/4.4
+        - temp/4.3
+    - url: https://github.com/couchbase/docs-sdk-php
+      branches:
+        - temp/4.3
+        - temp/4.2
+        - release/4.1
```
RichardSmedley, line 262: PHP 4.1 has just gone EOL - I'll be removing from the main playbook later.

osfameron: great, that should update automagically the next time I run the script...

PR #873, opened 2026-01-13
entry point for app services

iamfortune:


osfameron (changes requested):
accidental change to package-lock.json 😱

PR #880, opened 2026-03-11
DOC-14109 worktree support

osfameron:
This seems to work for me for simple cases now, but I feel like I'm missing:

* checking if current working tree is a linked worktree.
* the `worktrees: true` key

### DIFF
--- a/lib/preview.js
+++ b/lib/preview.js
@@ -80,6 +80,34 @@ module.exports.register = function () {
         }
     }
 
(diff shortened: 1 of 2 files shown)


osfameron:
Has been tested by @harini16r who also uses Worktrees, and solves her specific issue.
As it doesn't seem to introduce any regressions from my (non-worktree) testing, I'll merge, and see if it throws up anything else.

PR #889, opened 2026-06-09
Moving operator docs

TimLFletcher:
With the June release we need to make AI Services (Now AI Data Plane) a top level category.

I've been meaning to move Operator docs under Server for a while but I think now is the most needful and opportune moment.

Tested locally.


osfameron (approved):
Yup, it should be this simple 👍

PR #890, opened 2026-06-18
[AV-133854] AI Services Rename (ALL June releases playbook updates)

sarahlwelton:
Adding the attributes and updating landing pages to keep things going after AI Services rename. 

@osfameron, please shout if we need other playbooks updated than what I've grabbed here. I figured preview and the base was what mattered.


home/modules/ROOT/pages/cloud.adoc
```diff
@@ -18,7 +18,7 @@ Get access to SQL-like querying, full-text and vector search, powerful eventing,
 Capella Analytics is an analytical database (RT-OLAP) for real-time apps and operational intelligence.
 Capella Analytics is a standalone, cloud-only offering from Couchbase under the Capella family of products.
 
-Capella AI Services is a fully managed set of tools that help you build, deploy, and scale your agentic and retrieval-augmented generation (RAG) AI applications.
+The {ai-long} is a fully managed set of tools that help you build, deploy, and scale your agentic and retrieval-augmented generation (RAG) AI applications.
```
osfameron, line 21: Just a thought: the fact that all of these have "The" (or "the") preceding them suggests that that's really part of the name. 

E.g. if in future the product gets renamed to something that *doesn't* have a definite article, then this won't really help.

(Not a big issue, I think I'm coming around to this kind of attribute being an antipattern, personally, but that's not a team consensus, so just an FYI really 😂)

sarahlwelton: Not all of them do - see https://github.com/couchbaselabs/docs-ai/pull/86/files. 

We need the flexibility. The "the" cannot be a part of the name.

osfameron: Gotcha, but also I'm not explaining myself well cos still 🥴

It's more about an "ick" feeling that this kind of attribute makes you have to think in 2 places, to remember exactly how you've spelt the name, and then figure out whether you need "the" (or "The" or "a" or "A" or "An" or "an" etc.) and I don't really see the benefit over just writing the words in the original file, and editing them later when they need changing.

(As I said, not important, just struggling to explain what's going on in 🧠... Happy to approve the PR, as-and-when needed for the release 👍)

sarahlwelton: The benefits of using attributes for our product names aren't really just restricted to "update in one place and not worry about it" - happy to explain the reasoning if needed. Updating the modifiers is just a necessary evil.

osfameron: Sure thing: As I seem to be developing capital O "Opinions" about this, it's super useful if I'm missing something!

Why I've reached for this before personally has been:

1. to allow us to easily change naming when PM and Marketing aren't yet fully decided on it
2. to ensure Consistency in stylisation (spelling, capitalisation, etc.)

I do think that 2. is useful, especially for complex formatting (bold / superscript / unicode characters / stuff that Asciidoc finds hard to express like rare and exotic symbols like `++`... :badpokerface: etc.) but it does feel kinda icky - I guess maybe that's the "necessary evil" you mentioned 😂 🙈 

Sorry for sidetracking this PR review, can follow up on Slack 👍

zeba-shireen (approved):

PR #891, opened 2026-06-24
[DOC-14456] Add new connector versions

rakhi-prathap:
Updated the playbook to add new versions for connector docs - Power BI and Tableau.


osfameron (approved):
👍

sarahlwelton (approved):
🚢

PR #892, opened 2026-06-24
Doc 14466 llms txt update

TimLFletcher:
I'm assuming the attributes will work on the .md pages? Needs a sanity check.

### DIFF
--- a/home/modules/ROOT/partials/llms-txt.md
+++ b/home/modules/ROOT/partials/llms-txt.md
@@ -1,9 +1,13 @@
-> This is the official documentation for Couchbase, a leading NoSQL distributed database platform
-> designed for high performance, scalability, and flexibility.
-> The documentation covers comprehensive guides for developers and operations
-> teams, including installation and configuration, development tutorials, API
-> references, administration, security, cloud deployment, and integration with
-> various  programming languages and frameworks.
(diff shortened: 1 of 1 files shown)


osfameron (changes requested):
Nope, it's just a markdown chunk which is included verbatim in the output.

I suggest just writing the name in plain text tbh.

(Now, the creation of the llms.txt file is done in an Antora module, which does mean that we *can* have access to attributes, and *add* an attribute substitution pass.
I'd suggest the easier "writing words" workaround for this case, but especially if we're going to want to have more attributes in this llms.txt, then we could add this, let me know if you think it's worth it 👍)

TimLFletcher:
> Nope, it's just a markdown chunk which is included verbatim in the output.
> 
> I suggest just writing the name in plain text tbh.
> 
> (Now, the creation of the llms.txt file is done in an Antora module, which does mean that we _can_ have access to attributes, and _add_ an attribute substitution pass. I'd suggest the easier "writing words" workaround for this case, but especially if we're going to want to have more attributes in this llms.txt, then we could add this, let me know if you think it's worth it 👍)

Thanks, I did suspect this might be the case. It's not worth a change of functionality - I'll update the text.

osfameron (approved):
👍

PR #901, opened 2026-08-04
Installing node js and preview

pramodkumarka-sys:


osfameron (changes requested):
No, don't commit changes to package-lock.json, unless you have a good reason.
Installing the script should have run `npm ci` (a clean install, which uses and retains the package-lock)

`npm install` (which also updates the lock) should only be used when you *add* or *upgrade* a package.

Please go ahead and close this PR for now and let me know if you have any more questions 👍

pramodkumarka-sys:
Not intended PR. Resetting the changes.

